PR #3, opened 2019-11-11
feat: error integration

pmmmwh:
After a few nights of hard work this is finally presentable :tada:

There are still a lot of things that are missing here: no custom error overlays, no ways to disable the integration, no support for proxies, etc. Those will come later - I think it is more important for us to get the general DX right.

Please feel free to use this PR to track/iterate on changes that have to be made.

(Side note: I am debating to split the error overlay into a separate package/refactor with React, but didn't want to add build steps and mono-repo infra to this tiny plugin repo just yet)

@bvaughn @gaearon
Please test out the branch and comment on the DX 🙇

CC @charrondev @maisano - Since you two have adopted the concept early. I would be happy if you can try this out in real projects too.


gaearon:
Can you provide exact instructions on how to run/test this branch?

pmmmwh:
> Can you provide exact instructions on how to run/test this branch?

1. Git Clone the branch
2. Head to the example folder and run `yarn run start`

All plugin code is stored within the `src` folder so in case you want to run it against another project you can copy/paste it or use `yarn link` .

charrondev:
@pmmmwh FWIW the last time I was trying to use this I was unable to get it working w/ yarn link. We have a pretty complex webpack config though.

I had much better luck adding it through yarn pointing it to the github repo+branch.

I should have some time later today to test this out over on [vanilla/vanilla](https://github.com/vanilla/vanilla)

pmmmwh:
> @pmmmwh FWIW the last time I was trying to use this I was unable to get it working w/ yarn link. We have a pretty complex webpack config though.
> 
> I had much better luck adding it through yarn pointing it to the github repo+branch.
> 
> I should have some time later today to test this out over on [vanilla/vanilla](https://github.com/vanilla/vanilla)

I am not sure if this has anything to do with that, but I did fix a bug on using this branch as a sibling directory (i.e. `../plugin`). It used to recursively require itself, thus making Webpack unhappy.

charrondev:
Unfortunately it doesn't work at all for me because we have a custom webpack dev server config w/ a different public path than the site is served from.

Eg. Our local dev server is resolved at `http://dev.vanilla.localhost`

The actual dev server content is accessible from `http://dev.vanilla.localhost/EVERY_THING_HERE_RESOLVES_LOCALHOST_80`. The socket connection isn't mapped out though.

I'll try again later today and see if just proxying the socket over w/ my local nginx will work.

pmmmwh:
> Unfortunately it doesn't work at all for me because we have a custom webpack dev server config w/ a different public path than the site is served from.
> 
> Eg. Our local dev server is resolved at `http://dev.vanilla.localhost`
> 
> The actual dev server content is accessible from `http://dev.vanilla.localhost/EVERY_THING_HERE_RESOLVES_LOCALHOST_80`. The socket connection isn't mapped out though.
> 
> I'll try again later today and see if just proxying the socket over w/ my local nginx will work.

I am not sure if this branch at its current state could work with custom domains (because the URL resolving mechanism, from webpack `__resourxeQuery` is missing). I will check out the webpack config and see if I can get it working.

gaearon:
I'll just post the things I discover here as separate comments if that's okay.

gaearon:
Starting here:

```js
class ClassDefault extends React.Component {
  render() {
    return <h1>Default Export</h1>;
  }
}
```

Break it with a syntax error:

```js
import * as React from 'react';

class ClassDefault extends React.Component {
  render()
    return <h1>Default Export</h1>;
  }
}

export default ClassDefault;
```

Now change the code to introduce a runtime error *without fixing* the syntax error:

```js
import * as React from 'react';

class ClassDefault extends React.Component {
  render()
    throw new Error('nooo');
    return <h1>Default Export</h1>;
  }
}

export default ClassDefault;
```

Now fix the syntax error:

```js
import * as React from 'react';

class ClassDefault extends React.Component {
  render() {
    throw new Error('nooo');
    return <h1>Default Export</h1>;
  }
}

export default ClassDefault;
```

Expected: I see the runtime error
Actual: no runtime error until next save that changes the code.

gaearon:
## Module Init Error Not Shown

Start here:

```js
import * as React from 'react';

class ClassDefault extends React.Component {
  render() {
    return <h1>Default Export</h1>;
  }
}

export default ClassDefault;
```

Add a throw in module init phase:

```js
import * as React from 'react';

throw new Error('no')

class ClassDefault extends React.Component {
  render() {
    return <h1>Default Export</h1>;
  }
}

export default ClassDefault;
```

Expected: error overlay
Actual: error only in console

gaearon:
## What's Up With Babel Config?

Noticed there's a Babel config. 

```js
    // Applies the react-refresh Babel plugin on non-production modes only
    ...(!api.env('production') && { plugins: ['react-refresh/babel'] }),
```

Is it required? Should plugin error if it isn't on? Should plugin apply it automatically somehow? How to ensure user doesn't forget this or miconfigure this?

gaearon:
## Stuck Error

We start here.

```js
import * as React from 'react';

function FunctionDefault() {
  return <h1>Default Export Function</h1>;
}

export default FunctionDefault;
```

We add a new file. Let's call it `Foo.js`.

```js
export default function Foo() {
  return <h1>Foo</h1>
}
```

We edit our first file to use it.

```js
import * as React from 'react';
import Foo from './Foo';

function FunctionDefault() {
  return <Foo />;
}

export default FunctionDefault;
```

We get an error because `Foo` didn't import React. Fair. Let's add that to `Foo`.

```js
import * as React from 'react';

export default function Foo() {
  return <h1>Foo</h1>
}
```

Expected: this fixes the problem
Actual: I keep seeing the error overlay

gaearon:
## Tabbing and Focus

If I have

```js
setInterval(() => {
  throw Error('no')
}, 1000)
```

I can't tab into the close / next / previous buttons. Each re-render of the error overlay destroys them. Ideally it would keep the existing ones on the screen so keyboard users can still use it.

gaearon:
## Old Messages Showing Up

```js
let i = 0
setInterval(() => {
  i++
  throw Error('no ' + i)
}, 1000)
```

Let a few show up. Press "close".

Expected: The next one you'll see is `no N+1` if you last saw `no N`.
Actual: The next one you'll see is `no 1` again. Even though this one has been dismissed.

gaearon:
## Mobile Tap Area

I suspect that on mobile "prev" and "next" buttons have too little tap area. They need to be larger, at least on mobile.

gaearon:
## Wording and Emphasis

"Caught 1 error" implies it was *caught*. This can be misleading given that this actually shows uncaught errors. Also, the number isn't that important. You already see it below when there's more than one.

Let's instead put the actual error message at the top in red? That's the key part really.

gaearon:
## Overlay Design

We settled for minimal but maybe it's worth tweaking to get a bit closer to the design @rickhanlonii is building for React Native. Doesn't have to happen here, but this is a reference of what it roughly looks like.

Screenshot 2019-11-15 at 00 19 24

gaearon:
## Syntax > Runtime Errors

Start here.

```js
import * as React from 'react';

let i = 0
setInterval(() => {
  i++
  throw Error('no ' + i)
}, 1000)

export function FunctionNamed() {
  return <div />
}
```

Make a syntax error.

```js
import * as React from 'react';

let i = 0
setInterval(() => {
  i++
  throw Error('no ' + i)
}, 1000)

export function FunctionNamed() {
```

Expected: it stays forever until you fix it. After you fix it, you immediately see the last runtime error.
Actual: next runtime error makes the syntax error go away.

This is highly confusing because you may be trying to fix the *runtime* error, but your fix *won't work* as long as the syntax error exists. So it's crucial that syntax error always wins, when it exists.

gaearon:
## Boundaries

Wrap root in a boundary:

```js

class ErrorBoundary extends React.Component {
  constructor() {
    super()
    this.state = { hasError: false, error: null };    
  }
  static getDerivedStateFromError(error) {
    return {
      hasError: true,
      error
    };
  }
  render() {
    if (this.state.hasError) {
      return this.props.fallback;
    }
    return this.props.children;
  }
}

function App() {
  return (
    <ErrorBoundary fallback={<h2>error</h2>}>
      <FunctionDefault />
    </ErrorBoundary>
  );
}
```

Change the wrapped component to have a runtime error.

```js
function FunctionDefault() {
  throw new Error('no')
}
```

Expected: overlay shows up
Actual: no overlay until full reload

gaearon:
This is all I have so far. Good start, but needs more work. :-)

pmmmwh:
> This is all I have so far. Good start, but needs more work. :-)

Hey - Thanks for all the comments! It is really nice to receive clear and reproducible feedback on all the cases I missed.

I will reply them one by one, I think you wouldn't mind 😏

pmmmwh:
> ## What's Up With Babel Config?
> 
> Noticed there's a Babel config.
> 
> ```js
>     // Applies the react-refresh Babel plugin on non-production modes only
>     ...(!api.env('production') && { plugins: ['react-refresh/babel'] }),
> ```
> 
> Is it required? Should plugin error if it isn't on? Should plugin apply it automatically somehow? How to ensure user doesn't forget this or miconfigure this?

I have been thinking about this after seeing the work Parcel is doing to inject Babel plugins. Atm, I am thinking to have the plugin apply this automatically. If I have not mistaken, using multiple Babel loaders on the source code seems to be a do-able thing.

pmmmwh:
> ## Overlay Design
> 
> We settled for minimal but maybe it's worth tweaking to get a bit closer to the design @rickhanlonii is building for React Native. Doesn't have to happen here, but this is a reference of what it roughly looks like.
> 
> Screenshot 2019-11-15 at 00 19 24

On design - yea definitely. I am thinking to do a bit more on the overlay's side, presumably it could even be its own package that happens to be consumed by the plugin.

On functionalities, there are two things, Component Stacks and Source Tracing, that are pretty difficult to achieve (we discussed this on Twitter before). Component stacks will require React exposing them as an API, which I am not sure will happen, and Source Tracing will require specific types of source maps to be present (maybe show that as a conditional component, like when they are present?)

gaearon:
>Atm, I am thinking to have the plugin apply this automatically. 

Keep in mind the user's Babel wiring to Webpack might be more complicated than just a loader. E.g. maybe they use `happypack` or `thread-loader` or something else that adds indirection. So I don't know if it's actually feasible to inject it.

pmmmwh:
> > Atm, I am thinking to have the plugin apply this automatically.
> 
> Keep in mind the user's Babel wiring to Webpack might be more complicated than just a loader. E.g. maybe they use `happypack` or `thread-loader` or something else that adds indirection. So I don't know if it's actually feasible to inject it.

That's my initial concern on the injection. In fact, I am also not aware of ways to detect the plugin reliably too. I am thinking to add a mandatory instance of `babel-loader` which its sole use is to inject the plugin to all "JS-like" files, kinda like how the refresh runtime code was being injected. However, I am not aware of the implications of this approach, since I never thought using multiple instances of Babel is a good idea ...

gaearon:
I would suggest against it because it makes DEV builds slower (parse -> transform -> output twice) and user can't configure stuff like caching.

pmmmwh:
> I would suggest against it because it makes DEV builds slower (parse -> transform -> output twice) and user can't configure stuff like caching.

Yea.

> > Atm, I am thinking to have the plugin apply this automatically.
> 
> Keep in mind the user's Babel wiring to Webpack might be more complicated than just a loader. E.g. maybe they use `happypack` or `thread-loader` or something else that adds indirection. So I don't know if it's actually feasible to inject it.

Spent the night trying various ways to do this. If the user is using Webpack 4+ (which is the minimum version this plugin supports), and are not using something that makes Babel hidden from the plugin (e.g. `happypack` unfortunately, because how it works relies on calling the plugin interface and thus the bulk of transformation is happening outside of webpack), we can quite reliably detect if babel is used, and inject a plugin/a loader with the plugin conditionally.

Another way of doing it is to not inject the plugin directly, but to detect if the plugin have been applied. Since the plugin is solely related to function components (classes picked up by the plugin), we can scan source code after modules are built and detect traces of the plugin's work (e.g. `var _c;` and `$RefreshReg$` with the source code, along with something like `function` or `)\s*=>\s*{`. It should also be pretty reliable.

What do you think?

gaearon:
# Sourcemaps?

Noticed `cheap-module-source-map` is borked when the plugin is on. (And probably others.) Can't see original source or set breakpoints. Compare:

Screen Shot 2019-11-21 at 12 22 09 AM

Screen Shot 2019-11-21 at 12 23 47 AM

gaearon:
>we can scan source code after modules are built and detect traces of the plugin's work

Maybe you can inject a module that you'd *expect* the Babel plugin to "catch" with a RefreshReg call, and then verify that.

pmmmwh:
An update:

I have just pushed changes to solve all missed cases for error handling listed above. The design of the Error Overlay haven't been changed much, but some comments are also addressed. Still haven't landed a proper solution for Babel plugin detection as I am still experimenting with ways to do that. Source maps are on my radar, probably another difficult fix because of how they are handled in Webpack.

src/loader.js
```diff
@@ -10,9 +11,13 @@ const reactModule = /['"]react['"]/;
  * @returns {string} The injected module source code.
  */
 function RefreshHotLoader(source) {
+  // Add dependency to allow caching and invalidations
+  this.addDependency(path.resolve('./runtime/RefreshModuleRuntime'));
+
   // Only apply transform if the source code contains a React import
   return reactModule.test(source)
     ? source +
```
gaearon: I think this breaks source maps. You probably want to use the "source" abstraction instead which preserves mappings. See how webpack official loaders do it.

gaearon: I don't think the fix is difficult btw. I think you just need to use ConcatSource or something like this instead of actual concatenation.

pmmmwh: Can you test again with my latest patch? It should fix the issues with source maps. The issue was not properly accepting and passing source maps through.

pmmmwh: I had thought about using Dependencies, Sources and Templates like [this answer on StackOverflow](https://stackoverflow.com/a/52906440), but injecting external dependencies (`react-refresh`) this way was almost impossible without a full re-parse, so I went with a loader instead.

pmmmwh:
> > we can scan source code after modules are built and detect traces of the plugin's work
> 
> Maybe you can inject a module that you'd _expect_ the Babel plugin to "catch" with a RefreshReg call, and then verify that.

Implemented this idea as the default option with an escape hatch to disable the check just in case the user is doing something exotic.

gaearon:
Getting `Error: Cannot find module './runtime/TestComponent.js'` on this branch

pmmmwh:
> Getting `Error: Cannot find module './runtime/TestComponent.js'` on this branch

Fixed. I broke something after stashing, sorry.

gaearon:
Is it worth keeping it in a branch / unreleased at this point? Might be easier to test outside of this repo.

pmmmwh:
> Is it worth keeping it in a branch / unreleased at this point? Might be easier to test outside of this repo.

Maybe I'll merge this first, then release it under a scope so that the world can test it?

gaearon:
Sure

pmmmwh:
Found some last minute bug fixes and redid some parts of the overlay.

Time to merge this to release `v0.1.0` and watch the world crumble 🎉

Issue #15, opened 2019-12-10
BabelDetectComponent logic doesn't work for consumers


billyjanitsch:
First, thank you so much for your work on this project. I'm so excited to use it! 💙

While setting it up, I noticed what I think is an issue with the current logic around `BabelDetectComponent`.

The detection works by checking whether a "component" in `BabelDetectComponent.js` (part of the react-refresh-webpack-plugin aka RRWP runtime) was transformed by `react-refresh/babel`. Note that, when RRWP is installed by a consumer, `BabelDetectComponent.js` will end up in `node_modules/@pmmmwh/react-refresh-webpack-plugin/src/runtime/`.

The issue is that, in general, the `react-refresh/babel` transform shouldn't be applied to `node_modules/`. (For example, the [example in the readme](https://github.com/pmmmwh/react-refresh-webpack-plugin/tree/7a8dadafab4495e566f8edbc00e09d210637cd83#usage) applies the transform to a webpack rule that excludes `/node_modules/`.) Unless a RRWP consumer has their build configured incorrectly, `BabelDetectComponent.js` will never get transformed, and the "refresh check" check will always fail.

The example in this repo happens to work because it uses a [local version of the runtime](https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/7a8dadafab4495e566f8edbc00e09d210637cd83/example/webpack.config.js#L2) so the runtime modules don't match the loader exclusion rule.

For now, I can work around this by setting `disableRefreshCheck: true`.

billyjanitsch:
I suspect this is also the cause of https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/11.

pmmmwh:
> First, thank you so much for your work on this project. I'm so excited to use it! 💙
> 
> While setting it up, I noticed what I think is an issue with the current logic around `BabelDetectComponent`.
> 
> The detection works by checking whether a "component" in `BabelDetectComponent.js` (part of the react-refresh-webpack-plugin aka RRWP runtime) was transformed by `react-refresh/babel`. Note that, when RRWP is installed by a consumer, `BabelDetectComponent.js` will end up in `node_modules/@pmmmwh/react-refresh-webpack-plugin/src/runtime/`.
> 
[...]

Issue #11, opened 2019-12-07
The plugin is unable to detect transformed code from react-refresh


nthypes:
The plugin supports TypeScript (_.tsx files_)?

I'm getting the following error:

`✖ ｢wdm｣: Error: The plugin is unable to detect transformed code from react-refresh. Did you forget to include "react-refresh/babel" in your list of Babel plugins? Note: you can disable this check by setting "disableRefreshCheck: true".`

My **babel.config.js**:

```
module.exports = api => {
  // This caches the Babel config
  api.cache.using(() => process.env.NODE_ENV);
  return {
    presets: [
      "@babel/preset-env",
      "@babel/preset-react",
      "@babel/preset-typescript"
    ],
    // Applies the react-refresh Babel plugin on non-production modes only
    ...(!api.env("production") && { plugins: ["react-refresh/babel"] })
  };
};
```

My **webpack.config.js**:

```
const webpack = require("webpack");
const path = require("path");
const HtmlWebpackPlugin = require("html-webpack-plugin");
const ReactRefreshWebpackPlugin = require("@pmmmwh/react-refresh-webpack-plugin");

module.exports = {
  entry: "./src/index.tsx",
  output: {
    path: path.join(__dirname, "/dist"),
    filename: "main.min.js"
  },
  devServer: {
    hot: true
  },
  resolve: {
    extensions: [".ts", ".tsx", ".js"]
  },
  module: {
    rules: [
      {
        test: /\.(ts|tsx)$/,
        exclude: /node_modules/,
        use: [
          {
            loader: "babel-loader"
          }
        ]
      }
    ]
  },

  plugins: [
    new ReactRefreshWebpackPlugin(),
    new HtmlWebpackPlugin({
      template: "./src/index.html"
    })
  ]
};

```

pmmmwh:
> The plugin supports TypeScript (_.tsx files_)?
> 
> I'm getting the following error:
> 
> `✖ ｢wdm｣: Error: The plugin is unable to detect transformed code from react-refresh. Did you forget to include "react-refresh/babel" in your list of Babel plugins? Note: you can disable this check by setting "disableRefreshCheck: true".`
> 
> My **babel.config.js**:
> 
> ```
> module.exports = api => {
>   // This caches the Babel config
>   api.cache.using(() => process.env.NODE_ENV);
>   return {
>     presets: [
>       "@babel/preset-env",
>       "@babel/preset-react",
[...]

PR #17, opened 2019-12-10
fix: allow proper CRA usage

### LINKED ISSUE #8: Server startup hangs forever (or a really long time?) if disableRefreshCheck=false
I have to set `disableRefreshCheck: true` in order to successfully start up the webpack-dev-server, otherwise the startup seems to hang (I waited minutes, but not hours).

This is observable both on an ejected CRA app patched using instructions from the README as well as an existing CRA app that uses react-app-rewired/customize-cra.

### LINKED ISSUE #14: Fails to work when integrated into create-react-app
The error I am seeing is. Please let me know if there's anything else you need.

```
createSocket.js:30 Uncaught TypeError: connection.onClose is not a function
    at createSocket (createSocket.js:30)
    at Object../node_modules/react-refresh-webpack-plugin/src/runtime/ErrorOverlayEntry.js (ErrorOverlayEntry.js:85)
    at __webpack_require__ (bootstrap:785)
    at fn (bootstrap:150)
    at Object.0 (theme.ts:23)
    at __webpack_require__ (bootstrap:785)
    at checkDeferredModules (bootstrap:45)
    at Array.webpackJsonpCallback [as push] (bootstrap:32)
    at main.chunk.js:1
```

```
❯ npm ls react-refresh react-refresh-webpack-plugin webpack-dev-server webpack
frontend@1.1.13 /Users/xxx/frontend
├── react-refresh@0.7.0
├── react-refresh-webpack-plugin@0.1.0
└─┬ react-scripts@3.2.0
  ├── webpack@4.41.0
  └── webpack-dev-server@3.2.1
```

Taken from https://github.com/facebook/react/issues/16604#issuecomment-561967793
```js
const { addBabelPlugin, addWebpackPlugin, override } = require('customize-cra');
const ReactRefreshPlugin = require('react-refresh-webpack-plugin');

/**
 * @param {ReactRefreshPluginOptions} [options] Options for react-refresh-plugin.
 */
const addReactRefresh = options =>
  override(
    process.env.NODE_ENV === 'development'
      ? addBabelPlugin('react-refresh/babel')
      : undefined,
    process.env.NODE_ENV === 'development'
      ? addWebpackPlugin(new ReactRefreshPlugin(options))
      : undefined
  );

/* config-overrides.js */
module.exports = override(addReactRefresh({ disableRefreshCheck: true }));
```

pmmmwh:
This fixes the issues surrounding the Babel transform detection when `node_modules` JS files are either externalized (processed with another Babel config) or completely non-processed (e.g. TS project) and also issues related to the fallback `SockJS` integration.

The plugin will now split out a proper warning instead of error out and thus halts compilation. I have yet to think of a proper way to detect Babel transforms without using multiple conditional `RegExp`/AST matches, but to unblock people from using the plugin I think we should get this shipped for now.

Fixes #8
Fixes #14 
Partially addresses #11 and #15 but would love to do better - discussion started in #15.


allocenx:
Nice. And when do we get a new release? I really need this plugin asap. Thank you for the amazing work.

pmmmwh:
> Nice. And when do we get a new release? I really need this plugin asap. Thank you for the amazing work.

Most likely tomorrow if I don't get occupied by other stuff.

allocenx:
> > Nice. And when do we get a new release? I really need this plugin asap. Thank you for the amazing work.
> 
> Most likely tomorrow if I don't get occupied by other stuff.

Thank you. Will wait.

gaearon:
Don't forget that if it gets too complicated, we can just punt on detection altogether. It doesn't have to be a huge unbounded pile of hacks.

pmmmwh:
> Don't forget that if it gets too complicated, we can just punt on detection altogether. It doesn't have to be a huge unbounded pile of hacks.

Yea definitely, but I think there are still stuff that I haven't tried if it will work, and I am feeling adventurous :rofl:

pmmmwh:
> > > Nice. And when do we get a new release? I really need this plugin asap. Thank you for the amazing work.
> > 
> > 
> > Most likely tomorrow if I don't get occupied by other stuff.
> 
> Thank you. Will wait.

Released. :tada:

allocenx:
> > > > Nice. And when do we get a new release? I really need this plugin asap. Thank you for the amazing work.
> > > 
> > > 
> > > Most likely tomorrow if I don't get occupied by other stuff.
> > 
> > 
> > Thank you. Will wait.
> 
> Released. 🎉

Uhm un release page i still see only the v0.1.0 version...

allocenx:
Still not working, now receiving just: 

> WARNING in React Refresh Plugin: The plugin is unable to detect transformed code from react-refresh. Did you forget to include "react-refresh/babel" in your list of Babel plugins? Note: you can disable this check by setting "disableRefreshCheck: true".


I 100% followed the exact guide and i still got that.

Rule for JS:

`exclude: /node_modules/,
  test: /\.js?$/,
  use: {
    loader: 'babel-loader',
    options: {
      presets: ["@babel/preset-env", "@babel/preset-react"],
      plugins: ["react-refresh/babel"]
    }
  }`

Added into plugins too:

`plugins.push(
          new ReactRefreshWebpackPlugin()
        )
`

Why is this still not working?

pmmmwh:
> Why is this still not working?

It is because the detection still don't work properly. Everything fixed was about CRA/Webpack crashing silently, but not related to the detection. I am actively trying out other solutions to reduce false negatives, but that requires a lot of trial and error. I know it is noisy, I will add a note in the README to indicate that it is broken.

Issue #24, opened 2019-12-13
Getting error in web workers


GuskiS:
I'm getting error when using with web workers:
`Uncaught ReferenceError: $RefreshReg$ is not defined`

pmmmwh:
Can you provide me some steps to reproduce? I'm not too familiar with Web Workers and don't have much context on what is broken.

ro-savage:
@pmmmwh - I haven't had much time to looking into this yet.

But I believe the issue is caused because workers don't have window. And at least with https://github.com/developit/workerize-loader and https://github.com/GoogleChromeLabs/worker-plugin it suggests changing your webpack config to use `this` to allow for Hot Module Replacement

```js
  output: {
    globalObject: "this"
  },
```

When you do this Webpack is throwing on

```
bootstrap:22 Uncaught ReferenceError: window is not defined
    at __webpack_require__ (bootstrap:22)
    at bootstrap:97
    at bootstrap:97
```

Which is lines (https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/615157e5daadc7d1979e717e3cfb45c7c101a154/src/helpers/createRefreshTemplate.js#L13-L15)

It appears if you reference `window` at all here webpack throws even if its to check if `window` exists.

Changing these lines to use `this` seems to work for me for both when I've changed the globalObject to `this` and to when I have left it as 'window'.

rndm2:
I am getting same error as author with version 0.2.0 
I have added component that used in worker to babel ignore

pmmmwh:
> I am getting same error as author with version 0.2.0
> I have added component that used in worker to babel ignore

Can you provide more info on your setup?

rndm2:
> > I am getting same error as author with version 0.2.0
> > I have added component that used in worker to babel ignore
> 
> Can you provide more info on your setup?

We are using worker-loader from webpack to load workers to DOM. 
```
      {
        test: /\.worker\.js$/,
        use: {
          loader: 'worker-loader',
          options: {
            fallback: true,
            name: 'scripts/[hash].worker.js',
            publicPath: '/',
          },
        },
      },
```
I don't know why, but only last of the components chain have this error. 
I added this component to babel ignore. 
```
    "development": {
      "ignore": [
        "**/Element.js",
        "^^^ This needed to avoid transform of Element.js in Web Worker with react-refresh/babel, possible in future it can be removed"
      ],
      "plugins": [
[...]

PR #29, opened 2019-12-20
Use safeThis function in cleanup phase instead of "window" to better support web workers

ro-savage:
This attempts to address #24 where using `window` in clean up was causing webpack to throw when using web workers and setting webpacks `config.globalObject` to `this` as is suggested by https://github.com/developit/workerize-loader and https://github.com/GoogleChromeLabs/worker-plugin

I tested this on both a standard CRA and my own app with Web Workers. Both were working with fast-refresh.  

It appears in both cases `this` correctly references window when there is a window, while for workers it references itself (and I believe the worker doesn't have `$RefreshSetup$` attached).

However, I wasn't sure in what cases cleanup is supposed to fire. So I wasn't able to test it was correctly firing still.

### DIFF
--- a/src/helpers/createRefreshTemplate.js
+++ b/src/helpers/createRefreshTemplate.js
@@ -9,9 +9,19 @@ const { Template } = require('webpack');
  */
 const beforeModule = `
 let cleanup = function NoOp() {};
+let check = function (it) {
+  return it && it.Math == Math && it;
+};
 
-if (window && window.$RefreshSetup$) {
-  cleanup = window.$RefreshSetup$(module.i);
+let safeThis =  
+  check(typeof globalThis == 'object' && globalThis) ||
+  check(typeof window == 'object' && window) ||
+  check(typeof self == 'object' && self) ||
+  check(typeof global == 'object' && global) ||
+  Function('return this')();
+
+if (safeThis && safeThis.$RefreshSetup$) {
+  cleanup = safeThis.$RefreshSetup$(module.i);
 }
 
 try {


pmmmwh:
I don't think this change is enough to get stuff working, nor I think this is the correct approach. Global `this` have a lot of nuances that have to be dealt with, especially the module template is being called directly. There are also a few places that calls to `window` directly, and I don't they will work for workers either. (e.g. `window.location.reload()`, which is unavailable within worker scopes)

To better support multiple environments (for example, electron/web workers), I think we should properly detect the Webpack option of the global object and use that instead. I am thinking to use `compiler.options.output.globalObject`, which defaults to `window`, and inject it to any places that uses it with `webpack.DefinePlugin`. The only exception this won't work is in the module template where we will need to inject that manually (but that is easy to do since it is just a string interpolation). After all these changes the plugin should work as expected other than places where a full reload is needed - I am not sure what APIs we can call in a worker scope to get that working, maybe you can enlighten me a bit?

ro-savage:
This issue with using `compiler.options.output.globalObject` is that it's optional, web workers work fine without that option (but break using this plugin either way).

Changing this to `self` or `this` avoids the crashing when using web workers (and should refer to `window` for the cleanup), and hot refresh works as expected both for React Components and for the workers themselves (at least using workerize with my setup). 

That is, I can update a component or my worker, and it reloads the new code without resetting state.

I believe a web worker will always throw on `window`.
E.g. if I add `console.log(window)` into a worker, it throws.

So when you are attaching 
```js
if (window && window.$RefreshSetup $) {
  cleanup = window.$RefreshSetup$(module.i);
}
```
to the top of every imported module, it throws when attached to the web worker.

Instead, you could have a check if something is a web worker first
```js
const isWebWorker = typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope
if (!isWebWorker && window && window.$RefreshSetup$) { 
 cleanup = window.$RefreshSetup$(module.i); 
}
```

This would mean there is no cleanup for the web worker, but I am not sure that anything is ever attached to a web worker since you check for `window` before attaching event listeners etc. And a quick check showed the web worker's `this` and `self` did not have `$RefreshSetup$`.

I believe `self` should return `WorkerGlobalScope` ([mdn](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope)) for the worker and `window` for the browser. WorkerGlobalScope also include `location`.

Therefore either updating anywhere you refer to window as `self` or adding a check to decide `window` or `self` should be used could be done.

When I did try using `self` instead of window everywhere in the plugin, everything appear to still work fine from a quick test. However the worker still did not have `$RefreshSetup$` attached. Although maybe I did something incorrect?

However, I am not sure all this is required because it seems web workers have no problem reloading without having attaching anything special (maybe Webpack handles it without issue because its outside the DOM?)

Hopefully this helps, I can create a minimal repo with workers using CRA if that is helpful?

SimenB:
Should try to use `globalThis` instead?

pmmmwh:
> When I did try using `self` instead of window everywhere in the plugin, everything appear to still work fine from a quick test. However the worker still did not have `$RefreshSetup$` attached. Although maybe I did something incorrect?

Oh. I misunderstood the problem. I think using globalThis is the perfect solution, but to be safe we might have to use a shim to support older browsers. (Maybe copy and paste from core-js/global)

> Hopefully this helps, I can create a minimal repo with workers using CRA if that is helpful?

It will definitely help to have a repo to demonstrate the concept as well as for me to write tests later.

evertbouw:
Ran into this, manually changed `/node_modules/@pmmmwh/react-refresh-webpack-plugin/src/helpers/createRefreshTemplate.js` to
```js
if (globalThis.window && globalThis.window.$RefreshSetup$) {
  cleanup = globalThis.window.$RefreshSetup$(module.i);
}
```
This made it work for me

bugzpodder:
@evertbouw's fix works for me.  I use CRA but doesn't use web workers, not sure why I ran into this.

moimikey:
be nice to add this... having the same issue.

ro-savage:
@pmmmwh - I have updated this to use globalThis as suggested by @SimenB & @evertbouw 

However, according to [mdn globalThis is not supported by IE or the pre-chromium edge](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis).

My understanding is `this` would work fine, because we still check for `window`. But will leave it for you to decide.

Let me know if you want me to make any other changes.

pmmmwh:
> However, according to [mdn globalThis is not supported by IE or the pre-chromium edge](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis).

I am aware of that. I've read some discussions on this topic before and I think for this plugin, the following snippet would be sufficient ([ref](https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/global.js)):

```
var check = function (it) {
  return it && it.Math == Math && it;
};

module.exports =
  check(typeof globalThis == 'object' && globalThis) ||
  check(typeof window == 'object' && window) ||
  check(typeof self == 'object' && self);
```

> My understanding is this would work fine, because we still check for window. But will leave it for you to decide.

I am a bit uncomfortable with `this` - I know it should work fine, but if using the context-aware global object is feasible I would prefer that instead.

pmmmwh:
@ro-savage I am hoping to get this released in the next version - would you mind updating the PR/provide me access to your forked repo so that I can do it myself?

ro-savage:
@pmmmwh - I've update this to use your suggest check for `this`.

However, I  wasn't really sure how to put it in a seperate file an import it given that it's just a string that get inserted into other files.

I've given you access to the repo incase you have a better way to do it.

Looking forward to the update, thanks for the great work!

ro-savage:
Also, I've tested this locally using web workers and it all worked. But you probably want to double check and test it yourself.

pmmmwh:
Released in v0.2.0 🎉

Issue #28, opened 2019-12-20
Suggestion: Don't display error iframe


fwh1990:
Sometimes, the Promise reject data which is expected, but this plugin report error message by create an iframe#react-refresh-overlay element to html.body. That's boring, because **I can see error message in Console**.

This behavior will interrupted my mind and I have to close iframe by hand.

pmmmwh:
> Sometimes, the Promise reject data which is expected, but this plugin report error message by create an iframe#react-refresh-overlay element to html.body. That's boring, because **I can see error message in Console**.
> 
> This behavior will interrupted my mind and I have to close iframe by hand.

I thought the overlay only shows up when there are unhandled promise rejections? That's an Error by definition, isn't it? Or are you saying that it shows up even when the rejection is handled?

fwh1990:
I didn't handle promise rejection, that's my purpose. Would you please add an option to disable this behavior?

fwh1990:
I just tested Promise.reject in RN 0.61, it's a warn display at the bottom of screen. You recognize it as error.

Whatever, I also want to have an option to disable showing error.

pmmmwh:
My thoughts on this:

- Unhandled promise rejection should be warned as it is being warned in the console anyways. I can tune it down to be an error, but I wouldn't want to allow someone to switch of unhandled rejections independently in the error overlay
- Realistically, I can allow an option to make the plugin not inject the error overlay runtime code, which will effectively disable the overlay. I can see this useful in a few ways, for example custom HMR implementations or custom error logic. (also would be useful for CRA since they have both of these)

CC @gaearon your thoughts?

fwh1990:
Any progress?

pmmmwh:
> Any progress?

I'm currently on vacation and don't have much free time to do anything yet. You can submit a PR if you want this done quicker.

mattfwood:
For cases where you need to ignore a thrown exception, you should have something in your user code that is catching it and doing something with it (you could also use an [Error Boundary](https://reactjs.org/docs/error-boundaries.html#introducing-error-boundaries)):

```js
try {
  const res = await api.post('data');
  return res.data;
} catch (error) {
  // if the promise is rejected
  console.log(error); // log out the error, but catching here prevents this from being an uncaught exception
  // do something to display error message
}
```

[...]

PR #33, opened 2020-01-10
Make error overlay configurable

mmhand123:
For #28. Makes the error overlay configurable. Added it as default off - not sure if this is the direction you want, happy to change either way!


pmmmwh:
> For #28. Makes the error overlay configurable. Added it as default off - not sure if this is the direction you want, happy to change either way!

Hi - thanks for the PR. I think having it as default on would be better since we also need to cater for people that is not using CRA, and having the overlay on would bring a better experience for them.

pmmmwh:
#44 Supersedes this.

PR #37, opened 2020-01-30
note to transform BabelDetectComponent.js

jayenashar:


pmmmwh:
I am thinking to revisit this logic or even remove that check because it is causing more problems that its initial intention. I'll keep this open before I make the final decision.

pmmmwh:
I am deprecating the Babel detection logic for now because it is causing quite a bit of problems, and fixing it seems like too much effort for too little gain.

Thanks for the PR though!

Issue #7, opened 2019-12-06
Try integrating into Create React App


gaearon:
I've tried integrating this into an ejected CRA project but got stuck (https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/6).

This gets me thinking: perhaps, the next milestone is to get it to a point where we can create a pull request for CRA? What's missing for that to work? cc @iansu

iansu:
I would love to get this into Create React App. If there's anything needed on our end (other than installing and using the plugin) please let me know.

Ericnr:
Would be great to see this being implemented without ejecting from CRA too, using Craco or React-App-Rewired.

gaearon:
Yes, that’s what I’m referring to by “creating a pull request for CRA” 🙂 that was the goal all along.

pmmmwh:
> I've tried integrating this into an ejected CRA project but got stuck (#6).
> 
> This gets me thinking: perhaps, the next milestone is to get it to a point where we can create a pull request for CRA? What's missing for that to work? cc @iansu

To my knowledge, most of the work have to be done here:
1. Allow custom error overlay and alter react error overlay/the API needed here
2. Fix assumptions when sniffing Babel/Webpack, since CRA does not use webpack's mode argument
3. Need to try come up with an integration with the Webpack socket situation (related #9 ), and also make this work with servers under proxy/custom paths)

gaearon:
Could Babel detection be failing because the injected module doesn’t have a path that matches files going though Babel? Like usually people limit it to their source directory.

pmmmwh:
> Could Babel detection be failing because the injected module doesn’t have a path that matches files going though Babel? Like usually people limit it to their source directory.

Yes exactly. People often ignore `node_modules` so the injected module does not get processed, thus yielding high percentage of false positives.

esetnik:
Based on [the suggestion by @drather19](https://github.com/facebook/react/issues/16604#issuecomment-561961608) I have published a customize-cra plugin to make it easier. See [esetnik/customize-cra-react-refresh](https://github.com/esetnik/customize-cra-react-refresh).

bugzpodder:
I tested this with mostly a CRA 3.3.0 build and got a few issues:
1. Got the same error in #29 and that fix worked for me.
2. The first change always refreshes the page for some reason.  subsequent changes doesn't.
3. After a bit I get "The development server has disconnected.
[...]

PR #41, opened 2020-02-29
Fix CRA error overlay clashing with fast-refresh overlay

charrondev:
Related https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/7

The 2 error overlays were clashing. What I've done is

- Forked the CRA webpack entry and removed all warning & error handling. I added that to the `src` directory of the this package.
- Update the CRA example to use the local dev server entry from the package.

Additionally I've added a README section on setting up this project with `create-react-app` in addition to the other setup instructions.

## Expected

If you clone the repo, run the CRA example, then create some import/syntax error

- The new fast-refresh error overlay appears.
- If you fix the error, the overlay disappears.

## Unresolved in this PR

- There was a webpack conflict with the CRA example. The latest version of CRA wants `webpack@4.41.5`, but this repo doesn't have any top-level lockfile and as a result brings in a different, newer version of webpack. Actually resolving this in a robust way would likely require some type of monorepo tooling to make the examples separate private packages with their own independant dependency tree. **In the meantime, I added an `.env` file to bypass this check.**
- In order for the file to be able to be modified locally, I used the local devServer entry instead of the one from the published package. Ideally the whole example would use the local package so that users without publish access on this repo could work with it. But when I try to modify the main entry to point to it, I get errors like this:
  ```sh
Module not found: You attempted to import /Users/adamcharron/git/react-refresh-webpack-plugin/src/runtime/utils.js which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.
  ```
  In order to resolve this, once again I think some sort of monorepo tooling would be required so that the full local package would end up symlinked into `node_modules` and imported per-usual. Using `yarn link` or `npm link` seems to work but there's not really a good way to keep that checked into source control without additional tooling like lerna or yarn workspaces. **In the meantime, I left the main refresh entry pointing to the published NPM module and just pointed the dev server entry over the local package.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ module.exports = api => {
 
 **Create React App**
 
-See [the CRA example]([./examples/cra-kitchen-sink]).
+See [the CRA example](./examples/cra-kitchen-sink).
 
 Setting this up mostly relies on
 
--- a/README.md
+++ b/README.md
@@ -102,6 +102,22 @@ module.exports = api => {
 };
 ```
 
+**Create React App**
+
+See [the CRA example]([./examples/cra-kitchen-sink]).
+
+Setting this up mostly relies on
+
+- [Setting up `react-app-rewired` and `customize-cra`](https://github.com/arackaf/customize-cra).
+- [Applying the modifications in `config-overrides.js`](./examples/cra-kitchen-sink/config-overrides.js).
+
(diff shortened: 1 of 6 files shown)


charrondev:
Closing for https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/44

PR #44, opened 2020-03-01
feat(overlay): allow full customization of the error overlay integration

### LINKED ISSUE #28: Suggestion: Don't display error iframe

pmmmwh:
# Background

It has been a while since the last release, and a lot of opinion have been brought up related to the error overlay integration. This PR aims to address most of them, whether it is to not use it, or use some tool-dependent ones.

# Proposed API

```ts
interface ErrorOverlayOptions {
  /* The module you want to use as an error overlay. */
  module: string;
  /* The module you want to use as an entry point to set up the error overlay. */
  entry?: string;
}
```

Read [README.md](https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/feat/customize-overlay/README.md) for a better explanation since there some caveats related to this approach.

Fixes #28
Partially addresses #7
Supersedes #33 #41
Related facebook/create-react-app#8582

CC @gaearon @blainekasten @ianschmitz @Timer

Would love your thoughts as framework authors since you are the main target for the verbose API. I can create a PR in `create-react-app` after this lands so that `react-error-overlay` can be made compatible.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,25 +1,11 @@
 const path = require('path');
 const webpack = require('webpack');
-const { createRefreshTemplate, injectRefreshEntry } = require('./helpers');
+const { createRefreshTemplate, injectRefreshEntry, validateOptions } = require('./helpers');
 const { refreshUtils } = require('./runtime/globals');
-
-/**
- * @typedef {Object} ReactRefreshPluginOptions
- * @property {boolean} [disableRefreshCheck] Disables detection of react-refresh's Babel plugin.
- * @property {boolean} [forceEnable] Enables the plugin forcefully.
- * @property {boolean} [useLegacyWDSSockets] Uses a custom SocketJS implementation for older versions of webpack-dev-server
- */
-
-/** @type {ReactRefreshPluginOptions} */
-const defaultOptions = {
-  disableRefreshCheck: false,
-  forceEnable: false,
-  useLegacyWDSSockets: false,
-};
 
 class ReactRefreshPlugin {
   /**
-   * @param {ReactRefreshPluginOptions} [options] Options for react-refresh-plugin.
+   * @param {import('./types').ReactRefreshPluginOptions} [options] Options for react-refresh-plugin.
    * @returns {void}
    */
   constructor(options) {
@@ -27,7 +13,7 @@ class ReactRefreshPlugin {
   }
 
   /**
-   * Applies the plugin
+   * Applies the plugin.
    * @param {import('webpack').Compiler} compiler A webpack compiler object.
    * @returns {void}
    */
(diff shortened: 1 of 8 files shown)


blainekasten:
@pmmmwh I think this proposal looks pretty solid. When I started using this project, I was surprised to see a custom built error screen and not reuse `react-error-overlay` which is already pretty common and stable.

Can you maybe explain a bit why we need to have our own custom overlay within this project?

pmmmwh:
> Can you maybe explain a bit why we need to have our own custom overlay within this project?

@blainekasten 
First - Dan expects a `react-refresh` integration to have an overlay integration so that errors/warnings are immediately obvious, so the overlay integration had to be done anyways. On to why we're rocking a custom one, this is mainly because `react-error-overlay` imposes quite a bit of concerns on users' Webpack set up, and I didn't want that to affect the usability of this plugin.

Some of those includes: the `cheap-module-source-map` `devtool`must be used (it is probably not the best one, and some would decide to not use a `devtool`) and the IDE/source-linking integration requires custom server middlewares (which is not feasible for a lot of people). I could have sent PRs to get these issues resolved (or gracefully fallback to something less powerful), but I didn't that to block development/release of the plugin, so in the end I decided to go custom.

ianschmitz:
It would be good to get feedback from @timer, as i know that he had some thoughts on what a next gen of `react-error-overlay` may look like.

ro-savage:
Would be good to just be able to turn it off. I personally prefer just seeing the errors in the console.

pmmmwh:
> Would be good to just be able to turn it off. I personally prefer just seeing the errors in the console.

With this API you can just pass `false` and it will be off.

charrondev:
I like it! @pmmmwh As I said in https://github.com/facebook/create-react-app/pull/8582#issuecomment-594299782 once this is wrapped it will probably make for a better CRA integration (even though I like the overlay in this plugin more than the CRA one).

larsbs:
Any news on this?

pmmmwh:
> Any news on this?

I am trying to get a bit more feedback, so no concrete eta yet, sorry. Hopefully this can be merged next week.

src/index.js
```diff
@@ -50,6 +36,7 @@ class ReactRefreshPlugin {
 
     // Inject refresh utilities to Webpack's global scope
     const providePlugin = new webpack.ProvidePlugin({
+      [errorOverlay]: this.options.overlay && require.resolve(this.options.overlay.module),
```
larsbs: There's an issue with line. Basically, this is resolving to `[errorOverlay]: false` so webpack is complaining that there's no module named false

image

If you remove the line, everything works fine (but of course you won't have the overlay)

pmmmwh: Yes, I noticed this. A fix is on its way.

Issue #34, opened 2020-01-18
Support typescript loader ?


KurganskySergey:
Does this plugin force me to use babel ? I'm developing react project while using only typescript loader in webpack. Can i use your awesome tool for my project or maybe you have plans to support other loaders ?

TrySound:
This is babel specific tool. You can add babel after typescript or use only babel to transpile your code.
https://github.com/facebook/react/tree/master/packages/react-refresh/src

KurganskySergey:
looks like i can do it this way:
```
module: {
		rules: [
			{
				test: /\.tsx?$/,
				exclude: /(node_modules|bower_components)/,
				use: {
					loader: 'babel-loader',					
				},
			},
			// all files with a `.ts` or `.tsx` extension will be handled by `ts-loader`
			{
				test: /\.tsx?$/,
				loader: 'awesome-typescript-loader',
			}
		],
	},
```

no .babelrc no other configs for babel.

KurganskySergey:
If i change the order of rules it doesn't work
i think rules are applied in reverse order to files

markovicdenis:
You could use babel to compile typescript and use tsc --noEmit to have the error reports.  Or you could use tsc to convert typescript to JS and use the compiled JS as input paths in webpack (which I am currently doing now, this also allows you to skip some babel plugins like class-properties, decorators etc, since the tsc handles those).

KurganskySergey:
I use awsome-typescript-loader and i belive it's more convenient then both usecases you've described
- no babel's multiple dependencies
- no trash files in project structure

atjeff:
@KurganskySergey Would you mind sharing more of your webpack / tsconfig? I am having trouble setting this up

KurganskySergey:
```
const path = require('path')
const { CheckerPlugin } = require('awesome-typescript-loader')
const HtmlWebpackPlugin = require('html-webpack-plugin')
const Dotenv = require('dotenv-webpack')
const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin')
// ... your other imports

// You can tie this to whatever mechanisms you are using to detect a development environment.
// For example, as shown here, is to tie that to `NODE_ENV` -
// Then if you run `NODE_ENV=production webpack`, the constant will be set to false.
const isDevelopment = process.env.NODE_ENV !== 'production'

module.exports = {
	mode: isDevelopment ? 'development' : 'production',
[...]

PR #46, opened 2020-03-04
New example project: TypeScript without Babel

zet7777:
Hi, I used the solution from the issue [Support typescript loader ? #34](https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/34) in my own project. It works very well. You asked for a PR. I've decided to help and created this example project. Configuration rules were simplified.


examples/ts-without-babel/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "ts-without-babel",
+  "version": "0.1.0",
+  "author": "Eugene Scherbinin",
```
pmmmwh: I was thinking - maybe we should remove the author/license field from these examples, seems like that makes more sense ...

examples/ts-without-babel/.gitignore
```diff
@@ -0,0 +1,23 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
pmmmwh: I don't think we need another gitignore here?

examples/ts-without-babel/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "outDir": "dist",
+    "target": "es2018",
+    "module": "esNext",
+    "moduleResolution": "node",
+    "lib": ["dom", "es2015", "es2016", "es2017", "es2018", "es2019", "esnext"],
```
pmmmwh:
```suggestion
    "lib": ["dom", "dom.iterable", "esnext"],
```

examples/ts-without-babel/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+    "sourceMap": true,
+    "jsx": "react",
+    "baseUrl": ".",
```
pmmmwh: I don't think we need to use `baseUrl`?

examples/ts-without-babel/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "outDir": "dist",
+    "target": "es2018",
```
pmmmwh: We probably want this to be `es5`/`es6` instead for compatibility?

examples/ts-without-babel/webpack.config.js
```diff
@@ -0,0 +1,40 @@
...
+        use: isDevelopment ? [
+          {
+            loader: 'babel-loader',
+            options: { plugins: ['react-refresh/babel'] }
+          },
+          'awesome-typescript-loader'
+        ]
+        : 'awesome-typescript-loader'
+      },
+    ]
```
pmmmwh:
```suggestion
        use: [
          isDevelopment && {
            loader: 'babel-loader',
            options: { plugins: ['react-refresh/babel'] }
          },
          'awesome-typescript-loader'
        ].filter(Boolean)
      },
    ]
```

examples/ts-without-babel/webpack.config.js
```diff
@@ -0,0 +1,40 @@
...
+
+module.exports = {
+  mode: isDevelopment ? 'development' : 'production',
+  entry: './src/index.tsx',
+
+  module: {
+    rules: [
+      {
+        test: /\.tsx?$/,
+        include: path.join(__dirname, 'src'),
```
pmmmwh: Probably use exclude `node_modules` instead of including `src`?

examples/ts-without-babel/webpack.config.js
```diff
@@ -0,0 +1,40 @@
...
+    new HtmlWebpackPlugin({
+      template: './public/index.html',
+      filename: './index.html',
+    }),
+    isDevelopment && new ReactRefreshPlugin({ disableRefreshCheck: true }),
+  ].filter(Boolean),
+
+  resolve: {
+    extensions: ['.js', '.ts', '.tsx'],
+    modules: ['.', 'node_modules']
```
pmmmwh: What's this for?

README.md
```diff
@@ -102,6 +102,8 @@ module.exports = api => {
 };
 ```
 
+If you use TypeScript as a transpiler (without Babel) see this [example project](https://github.com/pmmmwh/react-refresh-webpack-plugin/tree/master/examples/ts-without-babel).
```
pmmmwh:
```suggestion
If you are using TypeScript (instead of Babel) as a transpiler, you will still need to setup `babel-loader` to process your source code. Check out this [example project](https://github.com/pmmmwh/react-refresh-webpack-plugin/tree/master/examples/ts-without-babel) on how to set this up.
```

examples/ts-without-babel/tsconfig.json
zet7777: It's for direct `import Some from 'src/some.ts'`. I think it's a must. But we can remove it to clear up the example. Remove?

examples/ts-without-babel/webpack.config.js
zet7777: It's again for direct `import Some from 'src/some.ts`'.

examples/ts-without-babel/package.json
zet7777: I haven't found any info on requirement of author and license. `npm init -y` creates both, and author is empty. I removed them in my example and tested  - okay.

examples/ts-without-babel/.gitignore
zet7777: Deleted gitignore.

examples/ts-without-babel/tsconfig.json
zet7777: Applied.

examples/ts-without-babel/tsconfig.json
zet7777: Changed to es6, but I'm not sure. We use `"module": "esNext"` (dynamic import) anyway and es6 is outdated. Maybe it's okay for example project.

examples/ts-without-babel/webpack.config.js
zet7777: Yes, it's better for consistency - applied. But FYI personally I don't like this hacky style. It's webpack 4. Initially I used `module.exports = (env, argv) => { ... }` and `if (argv.mode === 'development')`. In this case NODE_ENV and cross-env are not needed.

examples/ts-without-babel/webpack.config.js
zet7777: Why? If our typescript only in src folder it's better. And the [doc] (https://webpack.js.org/configuration/) says _Try to avoid exclude and prefer include_.

examples/ts-without-babel/tsconfig.json
pmmmwh: > It's for absolute paths like `import Some from 'src/some.ts'`. I think it's a must. But we can remove it to make the example more clean. Remove?

I guess this is more of a preference thing - to me its sole use is to remove relative imports and some might not want that functionality. I would prefer to stick to the defaults for an example project.

examples/ts-without-babel/tsconfig.json
pmmmwh: For full IE11 compatibility we will need ES5 (or ES6 if we're aiming for major browser compatibility, I think this is a good thing to do), then it's about the module resolution target - ESNext (I've heard CommonJS + `moduleResolution: node` works, but I'm not sure). This is a bit complicated because we're using Webpack, and Webpack handles the dynamic import syntax for us instead of letting it bleed to userland - so module `esnext` works sorta independently with the target ...

examples/ts-without-babel/webpack.config.js
pmmmwh: I again think this is down to personal preference - using functions would be fine for me too, but just wanted to keep consistency between examples.

examples/ts-without-babel/webpack.config.js
pmmmwh: Argh yea. Lets keep this.

examples/ts-without-babel/webpack.config.js
pmmmwh: I think we better stick to the default behaviour ...

jiyinyiyong:
Tried in my pre-existing project(which uses `ts-loader` as well as `thread-loader`) and seems it's not refreshing components. Is there a reason `awesome-ts-loader` is used in this PR, how about `ts-loader`?

pmmmwh:
> Tried in my pre-existing project (which uses `ts-loader` as well as `thread-loader`) and seems it's not refreshing components.

Do you have Babel properly setup? Can you create a reproducible example for me to investigate? Does your output contain references to `__webpack_require__.$Refresh$`?

> Is there a reason `awesome-ts-loader` is used in this PR, how about `ts-loader`?

No particular reason. I read that `awesome-typescript-loader` have been archived, so I am willing to accept a PR to change the current example to use `ts-loader`.

jiyinyiyong:
still trying...

> I read that awesome-typescript-loader have been archived, so I am willing to accept a PR to change the current example to use ts-loader.

Tried the demo and added a PR. This demo is working with `ts-loader`, I upgraded package to latest version and it's still able to update the component.

> Does your output contain references to __webpack_require__.$Refresh$?

I can find `__webpack_require__.$Refresh` in my output js code. Tried editing a component, but not updating.

The project I tried is quite different from the example above... https://github.com/jimengio/ts-workflow I'll try again to see if there's anything I can find out.

pmmmwh:
> The [project I tried](https://github.com/jimengio/meson-form) is quite different from the example above... I'll try again to see if there's anything I can find out.

How are you setting it up?

jiyinyiyong:
> With the DLL plugin you have to add react-refresh/runtime to the entries list

Oh, it's working. Thanks. Guess we need a warning in such cases, no clue to notice that DLL plugin caused the problem.

pmmmwh:
> > With the DLL plugin you have to add react-refresh/runtime to the entries list
> 
> Oh, it's working. Thanks. Guess we need a warning in such cases, no clue to notice that DLL plugin caused the problem.

#189 does add some steps to troubleshoot this. A warning is probably not going to happen because this error originates from wrong **ORDERING** of execution - we need to inject before React is called, but there's currently no known ways to detect whether React had executed (in both compile-time and runtime).

PR #52, opened 2020-03-25
Add custom sockHost/sockPort/sockPath

### LINKED ISSUE #40: Allow custom ports for websockets
We run webpack through a rails application and a gem called [webpacker](https://github.com/rails/webpacker)

We there for run our webserver at one port, but our dev server on another port.

Because the websocket just uses `window.localtion.port` (https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/master/src/runtime/createSocket.js#L26) it tries to connect to our web server port, not our web socket port.

This could be fixed a few ways.

1) You could allow a customer port to be passed to createSocket
2) You could use the port that is specified in webpack-dev-server's config `sockPort` (https://webpack.js.org/configuration/dev-server/#devserversockport), with a fall back webpack-dev-servers `port` and finally a fall back to `window.localtion.port`

This may also solve #36

marcello3d:
Add `sockHost` and `sockPort` option so that `react-refresh-webpack-plugin` can be configured to work through a proxy.

Tested using [patch-package](https://www.npmjs.com/package/patch-package).

Fixes #40


pmmmwh:
Hey - thanks for this PR! Can you also add handling for `sockPath`?


Also - is it possible for you to fix the merge conflicts (or give me access to your repo so I can push changes there)?

marcello3d:
Merged the basic conflicts, but need to update the README and verify that it still works.

marcello3d:
@pmmmwh I fixed up this branch and did more testing. It seems to work OK, but I've noticed that after the error overlay pops up from a build error, it never goes away until I reload the page (even if I get a subsequent successful build).

pmmmwh:
> @pmmmwh I fixed up this branch and did more testing. It seems to work OK, but I've noticed that after the error overlay pops up from a build error, it never goes away until I reload the page (even if I get a subsequent successful build).

I've checked - that is related to another issue. I will merge this first because what I'm working on depends on this, and I will fix that issue in another PR.

pmmmwh:
> it never goes away until I reload the page (even if I get a subsequent successful build).

Fixed in #57

marcello3d:
@pmmmwh sweet! are either of these fixes on NPM at this point? I see a `0.3.0-beta.0` release but not sure what's in that

pmmmwh:
> @pmmmwh sweet! are either of these fixes on NPM at this point? I see a `0.3.0-beta.0` release but not sure what's in that

Not yet, but beta 1 will contain them. I'll publish when I'm done with some other fixes.

pmmmwh:
Released in `0.3.0-beta.1` 🎉

PR #58, opened 2020-04-01
fix: make plugin work in electron renderers

pmmmwh:
This PR removes the dependency on `react-dev-utils/formatWebpackMessages`, which depends on `chalk`. The issue with chalk was that it calls `process`, which will fail in electron environments.


MOZGIII:
I'm pretty sure the root cause is that there should be multiple bundles, per-environment: one for node side and one for browser side. And each has to contain only the necessary code, to avoid invoking the functionality that's not available under a certain environments.

PR #69, opened 2020-04-08
remove check for "react" import

### LINKED ISSUE #67: React module detection doesn't work unless explicitly importing 'react'
And not 'react/index.js' or whatever. That's fine, but what I'm doing within our codebase is:

1. Babel uses `__react_createElement__` as its JSX pragma and these get injected during `babel-loader`.
2. Webpack's `ProvidePlugin` sees an unbound `__react_createElement__` and imports `node_modules/react/index.js`. (I don't know when this happens; maybe during the module templating? It's after the loaders in any case.)

But the loader in this plugin is looking for `'react'` or `"react"` in my source code. It looks in between step 1 and 2, so it sees `__react_createElement__` but not the import of React from `ProvidePlugin`.

The reason I do this is to avoid the `import react from 'react';` that we don't explicitly use in our source code, and it lets me switch between React and Preact pretty easily. You could justifiably say "don't do that" and just call it an unsupported use case, and I acknowledge that I'm going off the beaten path a bit, so I wouldn't expect something like this to be supported out of the box. However, how would you feel about giving a little bit of control over how modules are detected? The current regex of `/['"]react['"]/` could be the default, but for more _bespoke_ situations I could pass `/\b__react_createElement__\b/` or similar into the plugin options. What do you think? I'm willing to submit a pull request if you're okay with this.

(My other option is to create a loader that looks for my JSX pragma and injects `'react';` as a directive or `/*'react'*/` as a comment. I'll do that just so I can proceed, but it feels a little weird.)

xandris:
Pretty straightforward change :)

Fixes #67

### DIFF
--- a/src/loader/index.js
+++ b/src/loader/index.js
@@ -5,9 +5,6 @@ const RefreshModuleRuntime = Template.getFunctionContent(require('./RefreshModul
   .replace(/^ {2}/gm, '')
   .replace(/\$RefreshUtils\$/g, refreshUtils);
 
-/** A token to match code statements similar to a React import. */
-const reactModule = /['"]react['"]/;
-
 /**
  * A simple Webpack loader to inject react-refresh HMR code into modules.
  *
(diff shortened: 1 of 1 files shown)


pmmmwh:
Thanks for the PR!

PR #91, opened 2020-05-14
fix: handle failure to parse error stack traces

### LINKED ISSUE #90: ErrorStackParser.parse can throw an error
This line can throw, which causes an infinite set of errors.

https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/master/src/overlay/components/RuntimeErrorStack.js#L31

Caused by this line in the library https://github.com/stacktracejs/error-stack-parser/blob/master/error-stack-parser.js#L35

Not sure if this is helpful, but I have an error from the Chrome browser when a video is paused.

```
Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause(). https://goo.gl/LdLk22
```

This error has a few properties, but stack is not one of them:

```javascript
code: 20
message: "The play() request was interrupted by a call to pause(). https://goo.gl/LdLk22"
name: "AbortError"
__proto__: DOMException
```

This can be fixed by catching the exception, though there may be a better way to render an error with no stack trace:

```javascript
let errorStacks;
try {
  errorStacks = ErrorStackParser.parse(props.error);
} catch (e) {
  errorStacks = [];
}
```

mgh:
closes #90

### DIFF
--- a/src/overlay/components/RuntimeErrorStack.js
+++ b/src/overlay/components/RuntimeErrorStack.js
@@ -28,7 +28,14 @@ function RuntimeErrorStack(document, root, props) {
   stackContainer.style.lineHeight = '1.3';
   stackContainer.style.whiteSpace = 'pre-wrap';
 
-  const errorStacks = ErrorStackParser.parse(props.error);
+  let errorStacks;
+  try {
+    errorStacks = ErrorStackParser.parse(props.error);
+  } catch (e) {
+    errorStacks = [];
+    stackContainer.innerHTML = 'No stack trace is available for this error!';
+  }
+
   for (let i = 0; i < Math.min(errorStacks.length, 10); i += 1) {
     const currentStack = errorStacks[i];
 


src/overlay/components/RuntimeErrorStack.js
```diff
@@ -28,7 +28,14 @@ function RuntimeErrorStack(document, root, props) {
   stackContainer.style.lineHeight = '1.3';
   stackContainer.style.whiteSpace = 'pre-wrap';
 
-  const errorStacks = ErrorStackParser.parse(props.error);
+  let errorStacks;
+  try {
+    errorStacks = ErrorStackParser.parse(props.error);
+  } catch (e) {
+    errorStacks = [];
+    stackContainer.innerHTML = 'No stack trace';
```
michaelgmiller1, line 36: might be good to put a JSON dump of props.error as well as e.message for better debug-ability?

mgh: The expected exception is `throw new Error('Cannot parse given Error object');
 ` from https://github.com/stacktracejs/error-stack-parser/blob/master/error-stack-parser.js#L35, which doesn't look particularly useful.  It's also not something an end-user would want to debug -- the error would mainly be useful to authors of this plugin.

michaelgmiller1: got it, fair enough! (I'm not an approver, just looking at pull requests for fun)

pmmmwh: I think 'No stack trace is available for this Error!' seems clearer?

Otherwise I think this is good to merge.

mgh: Updated the message

PR #98, opened 2020-05-21
fix: remove all ES6 syntax

malwilley:
I have the unfortunate need to support IE11, so I noticed some places that were using non-ES5 features. 

I've tested on v0.3.1 and can confirm that IE11 works with these changes (and the polyfill provided in the readme).

### DIFF
--- a/src/runtime/refreshUtils.js
+++ b/src/runtime/refreshUtils.js
@@ -27,7 +27,7 @@ function getReactRefreshBoundarySignature(moduleExports) {
     return signature;
   }
 
-  for (const key in moduleExports) {
+  for (let key in moduleExports) {
     if (key === '__esModule') {
       continue;
     }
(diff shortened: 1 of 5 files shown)


src/runtime/refreshUtils.js
```diff
@@ -27,7 +27,7 @@ function getReactRefreshBoundarySignature(moduleExports) {
     return signature;
   }
 
-  for (const key in moduleExports) {
+  for (let key in moduleExports) {
```
pmmmwh, line 30: I think there's no need to switch to `let` from `const` since IE11 supports both?

malwilley: That was my thought as well! However, it appears that IE11 has one small caveat: https://caniuse.com/#feat=const

> Not supported in for-in and for-of loops

😞

pmmmwh: OH MY GOD IE WHY 🤦

malwilley:
Thanks for the swift response!

pmmmwh:
`v0.3.2` published.

Issue #79, opened 2020-04-22
How to integrate this into non-browser environments?


shirakaba:
I tried integrating this for my custom React renderer which uses a JS environment more similar to Node (e.g. uses `global` rather than `window`). After I aliased `global` as `window`, it got stumped by the lack of an `addEventListener` API:

```
file: node_modules/@pmmmwh/react-refresh-webpack-plugin/src/runtime/errorEventHandlers.js:85:10: JS ERROR TypeError: global.addEventListener is not a function. (In 'global.addEventListener(eventType, eventHandler)', 'global.addEventListener' is undefined)
```

What are the browser dependencies, and can anything be done to decouple it from the browser?

pmmmwh:
> What are the browser dependencies, and can anything be done to decouple it from the browser?

As of now, the core of plugin depends only on the location API (`window.location`) for reload functionality (and socket routing, if you're using WDS). The error overlay integration further depends on `window.addEventListener` and `window.removeEventListener` (and DOM APIs for rendering), because we listens on the `error` and the `unhandledrejection` events.

I'm not sure about your use case, but in reality you could disable the error overlay integration or use a custom error overlay setup to correctly wire up error-like events listening (only on 0.3.0). For the part related to the location API, I'm working on a feature to not force reload when unaccepted updates happen, so after that lands we should be somehow decoupled from the browser.

pmmmwh:
> Can anything be done to decouple it from the browser?

@shirakaba 
I've read a bit on the Node.js process API and it seems that the `uncaughtExceptionMonitor` event would cover what an error overlay requires. However, I'm not sure if that's something that is generally available for custom JS-based renderers. Would like your thoughts before I take any further action.

dangreen:
@pmmmwh in dev environment with ssr this plugin cause the error `ReferenceError: window is not defined`

pmmmwh:
> @pmmmwh in dev environment with ssr this plugin cause the error `ReferenceError: window is not defined`

Can you provide more background information, like where the error originates, whether the plugin is applied for the server bundle, etc.?

Also, see #36 for some solutions others have figured out.

shirakaba:
@pmmmwh Sorry for disappearing! Another project has taken up my full attention for the last month, but it should be finished in a few weeks, after which time I can dedicate more attention back to this.

Thank you very much for responding with details. I'll answer a few things now:

[...]

Issue #36, opened 2020-01-22
Issues working with Webpack Dev Middleware in server rendering environments


manuelbieh:
Hey,

I'm not sure if this has to do with this Webpack plugin or with Webpack itself, with Webpack Dev Middleware and in general, I'm not even sure if this makes sense at all. 

I've just spent two hours to get this really great plugin working in a [custom Server Side Rendering](https://github.com/manuelbieh/react-ssr-setup) environment that's currently using Webpack Dev Middleware (WDM) as Webpack's dev server and Webpack Hot Middleware (WHM) for HMR.

The issues I encountered:
- While my actual server rendered application is served on `localhost:8500`, my WDM is running on port 8501, but the socket connection is always created for  [`window.location.port`](https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/master/src/runtime/createSocket.js#L26) and is not configurable (yet)
- After I manually changed the port to 8501 in `runtime/createSocket.js`, it's still not working. There's an outgoing request to `http://localhost:8501/sockjs-node/info?t=1579726105901` but it looks like WDM isn't listening to a `/socksjs-node/info` route: `Cannot GET /sockjs-node/info`
- `POST /sockjs-node/213/kxx15upe/xhr_streaming` also gives me a 404
- Universal components that are also used on the server side fail because `$RefreshSig$ is not defined`, which seems obvious because they are [only defined when we're in a browser environment](https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/master/src/runtime/ReactRefreshEntry.js#L1-L13).:
```
ReferenceError: $RefreshSig$ is not defined
    at eval (webpack:///./src/shared/i18n/IntlProvider.tsx?:17:10)
    at Module../src/shared/i18n/IntlProvider.tsx (build/server/server.js:5202:1)
  ```

The last one makes sense but maybe we could just `assign () => {}` noops to `global.$RefreshSig$` and `global.$RefreshReg$`?!

After all, I'm not really sure if this plugin does even make sense in universal rendering scenarios. If it does, I'd be happy to get some helpful hints how this could be integrated into my certainly little bit unusual but probably not *that* exotic setup.

// edit:
current state of my implementation attempt can be found here:
https://github.com/manuelbieh/react-ssr-setup/compare/feature/react-refresh

Ephem:
[...]

PR #102, opened 2020-05-29
feat: set react-refresh globals on Webpack require instead of global scope

### LINKED ISSUE #88: Multiple entry points not working correctly
Our team has a config with the form:

```
{
  ...
  entry: {
    entry1: ['file1', 'file2'],
    entry2: ['file3', 'file4'],
  }
}
```

We were very excited to try out `react-refresh-webpack-plugin`, but unfortunately, could not get it working!

Through a process of trial and error, I removed `entry2`, and it started working. I think this is due to the logic in `injectRefreshEntry`, which would double-append react refresh into both entries. Is this intentional behavior?

Thanks!
Mike

pmmmwh:
This PR moves all react-refresh related globals to Webpack's require function instead of renderers' global scope.

This eliminates issues related to:
- Renderers that do not have `window` defined
- Electron/WebWorker/Native-related usage
- Multiple bundles included for the same page

Fixes #88.
Related to #36 and #79.

CC @sokra - if possible please take a look at this (possible issues/optimizations?)


michaelgmiller1:
Very excited for this!! If you need any help testing, happy to lend a hand!

manelio:
About the point

> Multiple bundles included for the same page

That fixes #88:

I modified the webpack-dev-server example this way:

webpack.config.js:
```
...
  entry: {
    main1: './src/index1.js',
    main2: './src/index2.js',
  },
...
```

public/index.html:
```
...
  <body>
    <div id="app1"></div>
    <div id="app2"></div>
  </body>
...

```

src/index1.js:
```
import * as React from 'react';
import { render } from 'react-dom';
import App1 from './App1';
render(<App1 />, document.getElementById('app1'));
```

src/App1.jsx:
```
import * as React from 'react';

function App1() {
  return <h1>Test1</h1>;
}

export default App1;

```

And similarly for index2.js and App2.jsx.

But the behavior is, if I made a change in App1.jsx, the DOM is updated, but nothing happens if I modify App2.jsx.

What does it mean "Multiple bundles included for the same page"? Is not this case?

pmmmwh:
> What does it mean "Multiple bundles included for the same page"? Is not this case?

Sorry I cannot dig into this right now, this might be a regression somewhere.

Can you try make both bundles share the same instance of React and React DOM and see if it works?

pmmmwh:
@manelio

Just add this to your Webpack config, it should work:

```js
module.exports = {
  // ... Other stuff
  optimization: {
    runtimeChunk: 'single',
  },
};
```

vaishnavi1029:
TypeError: $RefreshSig$ is not a function
I am writing tests and getting these error this when I am trying to check before and after context values change.
can you let me know why is this happening and how should I potentially eliminate it

pmmmwh:
You probably shouldn't apply `react-refresh/babel` when running tests.

PR #106, opened 2020-06-05
fix: do not use URLSearchParams to support IE 11 without polyfill

iiroj:
As mentioned in https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/88#issuecomment-633803815, we kept having issues with the top-level `URLSearchParams` call in `ErrorOverlayEntry.js`. Because it gets immediately invoked when the webpack chunk containing the code is loaded, it was running before our polyfill bundle and thus breaking IE 11.

Since the purpose of the offending code was to parse override options from the webpack `__resourceQuery`, I opened this PR to avoid the call entirely and instead just using `decoreURIComponent`, and `String.prototype.split`.

This should allow the plugin to work without polyfills on IE 11.

What do you think?

I'm not sure if the README has to be updated with regards to the polyfilll section.


pmmmwh:
Some background:

- `__resourceQuery` parsing is exclusively used for `webpack-dev-server` only
- Currently, the socket implementation for `webpack-dev-server` will depend on `URL` (inferred from usage of `native-url`)

The two above combined will mean even with this change, IE is still imcompatible, unless we do a custom implementation of `url.format`. I can add a shim to this myself later.

All that said - I think we can move the `__resourceQuery` parsing to `WDSSocket.js`.

iiroj:
Thanks for the comment. I looked at the usage of native-url, since we do use webpack-dev-server and the websocket implementation (instead of the default sockjs). It seems that part of the code then works fine on IE 11, at least in our case.

All in all, with this change it solved our problems.

I'm fine with it being solved another way though!

iiroj:
@pmmmwh I converted this PR into a draft to indicate it's not ready to be merged.

If I understood you correctly, I pushed a few further refactoring commits to move the parsing into a util method that's called internally in `WDSSocket.js`.

If you rather wish to take care of this yourself, let me know!

pmmmwh:
> If I understood you correctly, I pushed a few further refactoring commits to move the parsing into a util method that's called internally in `WDSSocket.js`.

Yea, this is how I've envisioned it. I'll think about chunking order issues.

I think moving forward maybe we'll have to have logic to make sure the error overlay is injected once only, like how the refresh hook should only be injected once. This can be handled later.

iiroj:
Hello,

I rebased this PR to the new `main` branch, and removed the draft status, since I had some time to test that everything is still working in IE 11 (and others).

iiroj:
I also opted to refactor the `parseURLSearchParams` into `getResourceQuery` that instead of being generic, only returns the parsed `__resourceQuery` global variable. I also added some tests for it.

pmmmwh:
Thanks for the PR! This will land in the next beta.

PR #116, opened 2020-06-15
fix: handle dependency clashes with injected modules

pmmmwh:
In #79 a bug was found where `__react_refresh_utils__` is `undefined` when using custom error overlay entries. It seems that we've hit a limitation of `webpack.ProvidePlugin` - injected modules will not contain references to other modules that are injected together.

This PR fixes that by excluding these modules from loader processing - one caveat is user-provided relative imports within `options.overlay.module` will also have to be excluded due to how module resolution works. This is a bit of a workaround but since the API is more of an advanced feature I think we can further polish it later.

Loader source map generation (when the loader is first in the chain) is also fixed here.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,7 @@
   },
   "overrides": [
     {
-      "files": ["src/{overlay,runtime}/**/*.js"],
+      "files": ["src/{overlay,runtime}/**/*.js", "src/loader/*.runtime.js"],
       "parserOptions": {
         "ecmaVersion": 2015
       },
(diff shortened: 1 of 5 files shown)


pmmmwh:
CC @sokra -
I'm really not sure if this is solvable gracefully.
It seems that modules injected via `ProvidePlugin` (or the parser in general) cannot contain any imports not from `node_modules`.

Stuff I've tried that did not work:

- Moving `refreshUtils` to `__webpack_require__.$Refresh$.utils` and lazily require it for each file
- Explicitly requiring `refreshUtils` in the injected runtime code via the loader
- Shuffling the order of modules in `ProvidePlugin`

As a last resort I can move everything from `refreshUtils` to the runtime ...

PR #121, opened 2020-06-18
Shorthand property names not supported in ie11

bmdoherty:
noticed this when using ie 11, first PR so let me know if it is ok


pmmmwh:
Thanks!

Issue #126, opened 2020-06-24
UnhandledPromiseRejectionWarning: Error: Cannot find module 'webpack/lib/ParserHelpers'


ckken:
(node:50522) UnhandledPromiseRejectionWarning: Error: Cannot find module 'webpack/lib/ParserHelpers'
Require stack:
- /Users/ken/Desktop/develop/MicroFE/emp-workspace/node_modules/@pmmmwh/react-refresh-webpack-plugin/src/index.js
- /Users/ken/Desktop/develop/MicroFE/emp-workspace/packages/emp-cli/webpack/config/plugin.js
- /Users/ken/Desktop/develop/MicroFE/emp-workspace/packages/emp-cli/webpack/config/common.js
- /Users/ken/Desktop/develop/MicroFE/emp-workspace/packages/emp-cli/helpers/project.js
- /Users/ken/Desktop/develop/MicroFE/emp-workspace/packages/emp-cli/scripts/dev.js
- /Users/ken/Desktop/develop/MicroFE/emp-workspace/packages/emp-cli/bin/emp.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:794:15)
    at Function.Module._load (internal/modules/cjs/loader.js:687:27)
    at Module.require (internal/modules/cjs/loader.js:849:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/Users/ken/Desktop/develop/MicroFE/emp-workspace/node_modules/@pmmmwh/react-refresh-webpack-plugin/src/index.js:6:23)
    at Module._compile (internal/modules/cjs/loader.js:956:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)
    at Module.load (internal/modules/cjs/loader.js:812:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Module.require (internal/modules/cjs/loader.js:849:19)
(node:50522) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:50522) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

pmmmwh:
Hi, can you please state the version of:

1. the plugin;
2. Webpack;


[...]

PR #130, opened 2020-06-28
fix: correctly detect hmr plugin in run and watch modes

pmmmwh:
This is related to #126 - the reason the plugin did not fail earlier is because of broken HMR detection in Webpack's watch mode.

This PR moves the detection down a step from `beforeRun` to `beforeCompile`, which will be called for both run and watch modes.


pmmmwh:
The PR has been merged but GitHub is bugging out 😃

PR #133, opened 2020-07-06
feat: revamp socket URL handling with currentScript source

### LINKED ISSUE #30: Get websocket port from `document.currentScript` if possible
Sorry if this is a duplicate. I saw there was #10 but I think this is a little different. With `webpack-dev-server` we're running it off another port locally. I looked at how `wds` tries to establish the socket and noticed `__resourceQuery`, but also that it falls back to trying to use `document.currentScript`. For the time being would it be ok to add in trying to parse `document.currentScript` and fall back to the current behavior? Happy to PR this, just wanted to check if it was something you'd want 😁

### LINKED ISSUE #105: iframe errors
image

These errors started after adding react-refresh and babel (typescript). any idea what causes this?

pmmmwh:
This fixes #105. It partially reverts changes in #106 because I was wrong about some assumptions on Webpack's `__resourceQuery`.

This change keeps IE11 compatibility for everyone other than WDS users (This is because, even if they polyfill URL, we will always run before the polyfill - unconfirmed though). All code with usage of the `URL` constructor is self-contained within `WDSSocket`.

Edit:
I've pulled the trigger to expand the scope of this.

Additional features resolves:
Fixes #30 
Fixes #100


pmmmwh:
Hmmmm this change will make further changes addressing resource query parsing much more difficult.

I was right about what I wanted to do - maybe I should just go ahead and handle #30 and #100 altogether ...

PR #159, opened 2020-07-24
lodash use of modular npm package

o-alexandrov:
Thank you for this plugin!
Use of modular npm packages of lodash [is deprecated](https://github.com/lodash/lodash/issues/3838) and will be removed in lodash v5.

### DIFF
--- a/overlay/index.js
+++ b/overlay/index.js
@@ -1,4 +1,4 @@
-const debounce = require('lodash.debounce');
+const { debounce } = require('lodash');
 const RuntimeErrorFooter = require('./components/RuntimeErrorFooter');
 const RuntimeErrorHeader = require('./components/RuntimeErrorHeader');
 const CompileErrorContainer = require('./containers/CompileErrorContainer');
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "ansi-html": "^0.0.7",
     "error-stack-parser": "^2.0.6",
     "html-entities": "^1.2.1",
-    "lodash.debounce": "^4.0.8",
+    "lodash": "^4.17.19",
     "native-url": "^0.2.6",
     "schema-utils": "^2.6.5",
     "source-map": "^0.7.3"
--- a/yarn.lock
+++ b/yarn.lock
@@ -6073,11 +6073,6 @@ lodash.camelcase@^4.3.0:
   resolved "https://registry.yarnpkg.com/lodash.camelcase/-/lodash.camelcase-4.3.0.tgz#b28aa6288a2b9fc651035c7711f65ab6190331a6"
   integrity sha1-soqmKIorn8ZRA1x3EfZathkDMaY=
 
-lodash.debounce@^4.0.8:
-  version "4.0.8"
-  resolved "https://registry.yarnpkg.com/lodash.debounce/-/lodash.debounce-4.0.8.tgz#82d79bff30a67c4005ffd5e2515300ad9ca4d7af"
-  integrity sha1-gteb/zCmfEAF/9XiUVMArZyk168=
-
 lodash.foreach@^4.5.0:
   version "4.5.0"
   resolved "https://registry.yarnpkg.com/lodash.foreach/-/lodash.foreach-4.5.0.tgz#1a6a35eace401280c7f06dddec35165ab27e3e53"


gaearon:
Let's just inline a simple implementation. Fewer dependencies is better to improve build times.

o-alexandrov:
Pretty much any app on the web has lodash in its dependencies (own or nested), why inlining (increasing maintenance complexity) is better?

What if you have another use of lodash API in the future, would you want to inline that also?

o-alexandrov:
More importantly this package as well as `react-refresh` both depend on `babel`.
- which in turn, depend on `lodash`

So including `lodash` in dependencies have no effect on build times.

gaearon:
Debouncing is literally five lines of code. Whether or not most apps use Lodash, it is unnecessary here.

>More importantly this package as well as react-refresh both depend on babel.

Babel is not a part of the bundled runtime. I believe you might be confusing packages that are used at the build time (which I'm not worried about) and the ones that get bundled as part of the development runtime. 

By the way, the assertion that these packages depend on Babel is incorrect. react-refresh itself has [no dependencies at all](https://unpkg.com/browse/react-refresh@0.8.3/package.json), and [this plugin](https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/main/package.json) only depends on it in development. Although this issue is beside the point since in either case Babel is used, of course.

TrySound:
Babel is trying to get rid of lodash eventually.
See here https://github.com/babel/babel/pulls?q=lodash

o-alexandrov:
> Babel is trying to get rid of lodash eventually.
This PR deals with the present, for anything else, an issue for long-term vision should be created.

@gaearon 
I haven't created this PR for wording games, you can see `yarn.lock` changes:
- it decreased

The `depend` word [above](https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/159#issuecomment-663567915):
- means this package needs `babel` to be installed & `react-refresh/babel` plugin to be part of the plugins added to babel config

That's why your claim is incorrect:
> Fewer dependencies is better to improve build times.

TrySound:
The main problem with your pull request is including whole lodash into user's browser. Using at least "lodash/debounce" solves the issue though better ditch lodash as very problematic and usually overused.

pmmmwh:
> The `depend` word [above](https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/159#issuecomment-663567915):
> 
>     * means this package needs `babel` to be installed & `react-refresh/babel` plugin to be part of the plugins added to babel config

I guess you and Dan are talking about two different aspects of the problem here. Dan is talking about `lodash` being in the final bundle (and hence parsed by the browser - i.e. **DEPEND**), and you're talking about `lodash` being installed on the disk and used during build time (i.e. **DEPEND**).

What we mean by keeping the runtime minimal is that we want to minimize the impact to the sipped bundle during development.

pmmmwh:
My thoughts on this are the same as @gaearon and @TrySound - inlining would be better since we do not depend on lodash in any where else. Whether the user uses `lodash` or not is out of our control, and we should optimize for the baseline - we should not penalize users that does not use `lodash` to have 70kB in their bundles just because we need that one tiny function.

gaearon:
```js
function debounce(fn, ms) {
  let timer;
  return function() {
    clearTimeout(timer);
    timer = setTimeout(fn, ms)
  }
}
```

o-alexandrov:
`lodash` would not be in the final bundle in `production` because of this plugin, it's an argument that has no ground.
This plugin is for development and wouldn't be even part of the `production` bundling process.

And when we talk about dead code removal in production using:
- modular lodash npm packages
is what harming tree shaking, as it is possible to simply alias `lodash` to `lodash-es` and many other alternatives (including [webpack 5's commonjs tree-shaking](https://github.com/webpack/changelog-v5#commonjs-tree-shaking), babel plugins for lodash, etc.), whereas libraries that implement modular npm package imports make tree-shaking a pain in production

With all respect, I'm surprised to see such comments from reasonably famous ppl in react community.

TrySound:
So you think bloating bundles in development is ok and will not harm anybody? lodash-es alias is one of problems I talked about. Setuping the package requires a lot of actions from developer (who may not even use it). It does not work properly out of the box.

gaearon:
@o-alexandrov 

My concern is not with the production environment. It is with development environment. I am speaking from experience with `react-error-overlay`, which we include by default in CRA. It is pretty bloated, which I was okay with, following the same train of thought are you're following now. But in practice this is not ideal.

For example, if you're testing how your app works under low network or CPU by using CPU or Network throttling, that extra development-only code is slowing down your testing so much it becomes very inconvenient. Of course, testing with production environment would be faster, but then you'd have to wait for each rebuild longer. The point is **there's absolutely no benefit to adding development-only 70 kB for a single five-line function**. Even if this code is development-only.

Overall, your attitude in this thread has been frustrating. I understand that you want to do good work but it would be nice if you didn't outright assume that you have all the context in the world, and other people's concerns are unfounded. You are repeatedly misunderstanding the argument.

pmmmwh:
> `lodash` would not be in the final bundle in `production` because of this plugin, it's an argument that has no ground.
> This plugin is for development and wouldn't be even part of the `production` bundling process.

Yes, but bundling also happens for development. Parsing and executing JavaScript has a cost, and we shouldn't ship code that we don't use even in development.
 
> And when we talk about dead code removal in production (we shouldn't, because this plugin has no relation to it) using:
> 
>     * modular lodash npm packages
>
> is what's harming tree shaking, as it is possible to simply alias `lodash` to `lodash-es` and many other alternatives (including [webpack 5's commonjs tree-shaking](https://github.com/webpack/changelog-v5#commonjs-tree-shaking), babel plugins for lodash, etc.), whereas libraries that implement modular npm package imports make tree-shaking a pain in production

Dead code removal does not necessarily only happen in production modes. For example, projects that use Google Closure Compiler often utilize it in development too to ensure that dead code elimination and code transformation won't be something that breaks production builds. With the ecosystem shifting to ESM, this will be more and more relevant because ESM tree shaking is guaranteed by the spec and can be done aggressively even in development to speed up compilation cycles (see Snowpack).

You mentioned that modular npm package imports are a pain for tree shaking with reference to Babel import plugins and Webpack aliasing, but for a library, it is somewhat an opposite story - we need to make sure that `lodash`'s debounce function is **NOT** tree shaken away, and to ensure that we optimize for users that have the most minimal setup.

Think about it in the most extreme case (actually, this pattern is quite common for some libraries e.g. `antd`'s notoriously big `icons` package) - what if the user alias `lodash` to something like this:

```js
import get from 'lodash/get';

export const get = get;
```

It is a side-effect and will break the whole plugin runtime.

pmmmwh:
I'll leave this PR open until we have another one that solves the issue that `lodash.debounce` is deprecated.

o-alexandrov:
@TrySound `lodash` is never bundled in development with a DllPlugin, as for the specific case, mentioned by Dan, there is another solution to his case.

---

On a side note, DllPlugin is probably used by most of the enterprise-level bundling processes in development.
With [snowpack](https://www.snowpack.dev/) arrival, there are even better solutions than DllPlugin for development.

---

@gaearon you have been pointing fingers, making your assumptions someone's competency twice in this thread and I have been talking on topic from the beginning, only stating facts. I agree that in the last message, I allowed myself to drop some disappointment in your judgement, but that was a reaction to your prior messages.

Coming back to the point, if you have some specific development needs, like developing for slow network, you have an option to use webpack's [DllPlugin](https://webpack.js.org/plugins/dll-plugin/) (which unfortunately doesn't work with react's hooks btw, but works with lodash), then caching individual DLL file during development, would exclude lodash & other libraries that you think slow down the development from loading.

PR #191, opened 2020-08-26
use ts-loader in typescript example, upgrade deps

jiyinyiyong:
As mentioned in https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/46#issuecomment-678902720 . `awesome-typescript-loader` has been archived. So `ts-loader`.

Tried demo and it's not breaking.

### DIFF
--- a/examples/typescript-without-babel/package.json
+++ b/examples/typescript-without-babel/package.json
@@ -7,16 +7,16 @@
     "react-dom": "^16.13.1"
   },
   "devDependencies": {
-    "@babel/core": "^7.10.5",
+    "@babel/core": "^7.11.4",
     "@pmmmwh/react-refresh-webpack-plugin": "^0.4.1",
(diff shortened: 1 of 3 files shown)


pmmmwh (approved):

pmmmwh:
Hi, can you pin the TypeScript version instead of using `^`? Also - it would be great if you can keep the dependencies list sorted.

jiyinyiyong:
Force pushed.

Why it's better to pin a specific version?

pmmmwh:
> Why it's better to pin a specific version?

TypeScript does not follow SemVer and to my understanding could ship breaking changes in so-called `minor` releases (I know the TS team minimize them).

For an example to illustrate that this plugin is working (sort of like e2e tests), it is better to pin it so that we eliminate TS as a variable.

PR #212, opened 2020-09-28
feat: make react-refresh singleton to support multiple webpack bundles

tianjianchn:
Make react-refresh singleton to support multiple webpack bundles' integeration, just like react-hot-loader.

It's usually used in the micro-frontend scenes like below:
1. There are some(>=2) frontend projects(repos), which each has webpack, react, react-dom, react-refresh
2. When user visits the page, there may be two or more projects run in the browser
3. Mostly, we would use singleton react, react-dom and so on, which means that they are shared between projects(repos).

So, without this pr, each project(repo) will use their own react-refresh instances, which result components not re-render after module hot updated.

### DIFF
--- a/client/ReactRefreshEntry.js
+++ b/client/ReactRefreshEntry.js
@@ -1,9 +1,14 @@
+/* global reactRefreshGlobal */
+
 const safeThis = require('./utils/safeThis');
 
 if (process.env.NODE_ENV !== 'production' && typeof safeThis !== 'undefined') {
   // Only inject the runtime if it hasn't been injected
   if (!safeThis.__reactRefreshInjected) {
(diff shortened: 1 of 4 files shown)


pmmmwh:
Hi, instead of doing this I think the better way is to have this in user-land.

For example you could share `react-refresh` just like how you would share `react`and `react-dom`.

You can probably also use `optimization.runtineChunk = 'single'`.

Apart from that, the core of this plugin is currently platform agnostic, so I would not want to use `window` and even global variables to achieve this. If we were to use global variables, we would have had done this from the beginning.

tianjianchn:
Agreed! By making `react-refresh/runtime` external in webpack and assigning to global(`window.ReactRefreshRuntime = require('react-refresh/runtime')`), all are also OK now.
Thanks for your response!

PR #234, opened 2020-10-25
ci: fix test failures

pmmmwh:
~This PR fixes Chromium for CI - sandboxing is now disabled due to difficulty to have it working reliably (If we had the option to run Docker with more privileges it would work but not at the moment).~

This PR fixes multiple stuff for CI:

- Authenticated Docker Pulls: since 2 Nov, 2020 Docker Hub have imposed rate limiting for image pulls, so auth is added to cope with that (higher rate limit)
- Using new batch of `cimg-puppeteer` under the hood now
- Added tests for Node.js 15 and bumped all LTS versions to latest security patches
- Bumped `webpack@latest` and also the corresponding CLI to latest versions


pmmmwh:
Scheduled publish of new batch of fixed image at 00:00 UTC - Blocked until that happens and all tests are re-triggered.

pmmmwh:
One last remaining issue is on swapping Webpack versions within a `child_process` ... 🤦🏻‍♂️ 
We might have to rethink how we swap Webpack 4/Webpack 5 versions right now ... sigh.

pmmmwh:
Tests now run properly on both Webpack 4 and 5 - and for Webpack 5 they are failing correctly since we haven't caught up to all latest changes from webpack 5.0 to 5.4 ...

pmmmwh:
Tests should pass now 🤞🏻

PR #236, opened 2020-10-27
fix handling dynamic path

### LINKED ISSUE #216: Fast refresh error - can't access property "trim", t is null
## Issue

In [this example repo](https://github.com/frehner/react-fast-refresh-externals), I get the following error 100% of the time:

`Uncaught (in promise) TypeError: can't access property "trim", t is null`

in debugging, it comes from this function 

https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/89ccca3a528a68f06f777daef09a28a5614e5525/sockets/utils/getCurrentScriptSource.js#L5

which does various things to try and figure out the current script source. However, in that example repo, `document.currentScript` is null AND `currentScript.getAttribute('src');` is null - which then causes issues down the line: specifically here (where the error is eventually thrown)

https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/89ccca3a528a68f06f777daef09a28a5614e5525/sockets/utils/getSocketUrlParts.js#L22

## Fix proposal

With that information, I propose the following change to the `getCurrentScriptSource` (with the caveat that I am not a webpack plugin developer, so I don't know the ins-and-outs of webpack plugins and this may not be the correct solution. However, I have tested this proposal and it appears to fix everything in that example repo)

```js
getCurrentScriptSource() {
  return __webpack_public_path__
}
```

which, again, fixes the issue for that example repo... and potentially fixes any issues for anyone using a [dynamic public path](https://webpack.js.org/guides/public-path/#on-the-fly) as well?

If you think that's the correct way to go, then I can make a PR for it. 

Thanks for this project, and I'm looking forward to your thoughts.

frehner:
fixes #216 

because getCurrentScriptSource executes at runtime in the browser, it can simply look at `__webpack_public_path__` which means that it should always be up-to-date, even if an app dynamically changes it. 

however, we have to allow the app to execute before the plugin; if we don't, and the app [dynamically sets the public path](https://webpack.js.org/guides/public-path/#on-the-fly) then it will have been too late for this plugin to see that change and it will attempt to connect to the wrong place. 

`yarn test` passes

`yarn test:webpack-5` fails, but that appears to happen on master as well, so it doesn't appear to have regressed from what I can tell

I'm not extremely familiar with webpack plugin development, so feel free to suggest changes or even just close this and use it as inspiration for a better update

### DIFF
--- a/lib/utils/injectRefreshEntry.js
+++ b/lib/utils/injectRefreshEntry.js
@@ -88,15 +88,14 @@ function injectRefreshEntry(originalEntry, options) {
   }
   // Multiple entry points
   if (typeof originalEntry === 'object') {
-    return Object.entries(originalEntry).reduce(
-      (acc, [curKey, curEntry]) => {
-        if(curKey === 'vendor') {
-          return {
-            ...acc,
-            [curKey]: [...prependEntries, ...overlayEntries, ...curEntry]
-          }
-        }
-        return ({
+    return Object.entries(originalEntry).reduce((acc, [curKey, curEntry]) => {
+      if (curKey === 'vendor') {
+        return {
+          ...acc,
+          [curKey]: [...prependEntries, ...overlayEntries, ...curEntry],
+        };
+      }
+      return {
         ...acc,
         [curKey]:
           typeof curEntry === 'object' && curEntry.import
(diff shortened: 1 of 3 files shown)


pmmmwh:
Hi - I'm a bit unsure on this as I think it will break the error recovery flow.

Can you try throwing errors in runtime and see if the error overlay shows up (and also if the build can recover after the error)?

frehner:
you're right, that didn't work. I'm currently investigating other methods of getting the socket connection.

PR #241, opened 2020-10-30
fix: adds resourceQuery to getSocketUrlParts to setup the socket

### LINKED ISSUE #216: Fast refresh error - can't access property "trim", t is null

frehner:
Fixes #216 

by default it pulls from WDS's devServer option in the webpack config, and passes that as a resourceQuery.

getSocketUrlParts changes allow it to use the resourceQuery primarily, and then fall back to the old method if resourceQuery isn't there.


lib/utils/injectRefreshEntry.js
```diff
@@ -29,11 +30,23 @@ function isSocketEntry(entry) {
...
  * @param {import('../types').NormalizedPluginOptions} options Configuration options for this plugin.
+ * @param {import('webpack').Compiler} compiler A webpack compiler object.
  * @returns {WebpackEntry} An injected entry object.
  */
-function injectRefreshEntry(originalEntry, options) {
+function injectRefreshEntry(originalEntry, options, compiler) {
   /** @type {Record<string, *>} */
   let resourceQuery = {};
+
+  if (compiler && compiler.options && compiler.options.devServer) {
```
patrikholcak, line 40: Not sure if depending on `options.devServer` is the best solution here, dev server could also be called from CLI or by `new WebpackDevServer()` in that case `options.devServer` might be undefined

lib/utils/injectRefreshEntry.js
```diff
@@ -29,11 +30,23 @@ function isSocketEntry(entry) {
...
+    if (compiler.options.devServer.sockHost)
+      resourceQuery.sockHost = compiler.options.devServer.sockHost;
+    if (compiler.options.devServer.sockPath)
+      resourceQuery.sockPath = compiler.options.devServer.sockPath;
+    if (compiler.options.devServer.sockPort)
+      resourceQuery.sockPath = compiler.options.devServer.sockPort;
+  }
+
   if (options.overlay) {
     options.overlay.sockHost && (resourceQuery.sockHost = options.overlay.sockHost);
```
patrikholcak, line 51: Shameless plug that solves my problem: Could we also add this here?

```suggestion
    options.overlay.sockProtocol && (resourceQuery.sockProtocol = options.overlay.sockProtocol);
    options.overlay.sockHost && (resourceQuery.sockHost = options.overlay.sockHost);
```

sockets/utils/getSocketUrlParts.js
```diff
@@ -18,36 +18,56 @@ const parseQuery = require('./parseQuery');
...
-  // which filters out falsy values when concatenating the `auth` string.
-  // In reality, we need to check for both values to correctly inject them.
-  // Ref: GoogleChromeLabs/native-url#32
-  // The placeholder `baseURL` is to allow parsing of relative paths,
-  // and will have no effect if `scriptSource` is a proper URL.
-  const authUrlParts = new URL(scriptSource, 'http://foo.bar');
-  // Parse authentication credentials in case we need them
-  if (authUrlParts.username) {
-    auth = authUrlParts.username;
+    urlParts = url.parse(
```
patrikholcak, line 33: See this: there already is a way to parse the `resourceQuery`

https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/3885807f741b97e5b744b2e6a9687aa0292b33eb/sockets/utils/getSocketUrlParts.js#L100-L103

Let’s remove it from the bottom so we don’t do it twice?

lib/utils/injectRefreshEntry.js
```diff
@@ -29,11 +30,23 @@ function isSocketEntry(entry) {
...
+ * @param {import('webpack').Compiler} compiler A webpack compiler object.
  * @returns {WebpackEntry} An injected entry object.
  */
-function injectRefreshEntry(originalEntry, options) {
+function injectRefreshEntry(originalEntry, options, compiler) {
   /** @type {Record<string, *>} */
   let resourceQuery = {};
+
+  if (compiler && compiler.options && compiler.options.devServer) {
+    resourceQuery.domain = createDomain(compiler.options);
```
patrikholcak, line 41: Shouldn’t this be 

```suggestion
    resourceQuery.domain = createDomain(compiler.options.devServer);
```

lib/utils/injectRefreshEntry.js
frehner: cli args get merged into the `devServer` object, so it appears safe to use this

(to test, I used `port:8080` in my devServer config, but then ran the cli with `--port 9090`. the `devServer` on this line had `port: 9090`

frehner: also, I think it's best to use whatever `devServer` defaults to, and then still keep the options for `options.overlay` to overwrite it when you want to have that extra flexibility, which I think my code here does (note a couple of lines down where options.overlay overwrites resourceQuery)

lib/utils/injectRefreshEntry.js
frehner: great catch! 👍

lib/utils/injectRefreshEntry.js
frehner: I'm open to adding it if @pmmmwh wants it

sockets/utils/getSocketUrlParts.js
frehner: agreed. thanks!

frehner:
updated. thanks for the review @patrikholcak

patrikholcak:
@frehner Could you please take a look at this commit? https://github.com/patrikholcak/react-refresh-webpack-plugin/commit/b60aac065f85394a7a0aa50bed9df2cf898fba27

I looked into it again and wanted to ask if this really fixes the problem

What I’m mainly talking about is this:
https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/71dacc6761516da8c972fc8d013ca8fd38fa00b8/sockets/utils/getSocketUrlParts.js#L33-L40

If you look into the tests in this repo, the `resourceQuery` for `getSocketUrlParts` doesn’t look like the one from webpack dev server which makes me believe your changes are incompatible. here it looks like this:

```
?sockHost=foo.com&sockPath=/socket&sockPort=9000
```

I tried fixing it and I don’t think the `getSocketUrlParts` needs that many changes. We probably want to leave as much of the original code as possible. eg. `url.parse('')` results in all keys being `null` and if you follow the code, the function does already handle `null` nicely (and I think that was the original intention here). The original code also handles default values nicely and that’s probably why the `parseQuery` was all the way at the bottom of the function.

I’m sure the `domain` option could just be an `if` statement inside of the `getSocketUrlParts` if that fixes some weird state for you.

I didn’t want to submit another PR with my changes because you already opened this one, but I can submit one to your repo and we can work in my changes if you want. And if not, that’s ok, too

lib/utils/injectRefreshEntry.js
patrikholcak: Turns out it’s not that trivial to add it. Opened a PR for it https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/242

pmmmwh:
Ah I think the failing tests are making this much harder to debug 🙈 
I'll try to squeeze some time to fix tests later today.

Meanwhile, I did a quick look at this PR and I think some of the changes here are not necessarily needed since for use our `resourceQuery` is really a `URLSearchParams` castable query string.

I think the changes proposed by @patrikholcak looks very promising to me - it would be really great if you two can consolidate the changes. Super thankful for you guys to explore this issue (my personal use cases for Webpack are usually not that complex 😆 )!

frehner:
> @frehner Could you please take a look at this commit? [patrikholcak@b60aac0](https://github.com/patrikholcak/react-refresh-webpack-plugin/commit/b60aac065f85394a7a0aa50bed9df2cf898fba27)
> 
> I looked into it again and wanted to ask if this really fixes the problem
> 
> What I’m mainly talking about is this:
> 
> https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/71dacc6761516da8c972fc8d013ca8fd38fa00b8/sockets/utils/getSocketUrlParts.js#L33-L40
> 
> If you look into the tests in this repo, the `resourceQuery` for `getSocketUrlParts` doesn’t look like the one from webpack dev server which makes me believe your changes are incompatible. here it looks like this:
> 
> ```
> ?sockHost=foo.com&sockPath=/socket&sockPort=9000
> ```
> 
> I tried fixing it and I don’t think the `getSocketUrlParts` needs that many changes after all. We probably want to leave as much of the original code as possible. eg. `url.parse('')` results in all keys being `null` and if you follow the code, the function does already handle `null` nicely (and I think that was the original intention here). The original code also handles default values nicely and that’s probably why the `parseQuery` was all the way at the bottom of the function.
> 
> I’m sure the `domain` option could just be an `if` statement inside of the `getSocketUrlParts` if that fixes some weird state for you.
> 
> I didn’t want to submit another PR with my changes because you already opened this one, but I can submit one to your repo and we can work in my changes if you want. And if not, that’s ok, too

hey that looks good. if you don't mind, go ahead and add a PR to my branch and we can definitely go with that simpler code. seems like it should work in my case I think - I'll verify it once it's merged. 👍

frehner:
@patrikholcak I updated the code and added you as a co-author to the commit to give you credit for the work - let me know if that's ok. 

@pmmmwh I've incorporated the feedback and this code looks like it fixes the issue with my setup, so I'm going to remove the "In Progress" status from this PR in hopes that we're close 🙂

patrikholcak:
@frehner I submitted a PR to your repo https://github.com/frehner/react-refresh-webpack-plugin/pull/1

I just need to check it one more time before I wanted to message you about it

frehner:
> @frehner I submitted a PR to your repo [frehner#1](https://github.com/frehner/react-refresh-webpack-plugin/pull/1)
> 
> I just need to check it one more time before I wanted to message you about it. Sorry, I didn’t have much time to work on it

my fault - I totally missed that you had opened a PR for it. 

but in any case, I just merged in the additional work that you added. it's looking good! thanks for all your work

patrikholcak:
🙌 great! pPleasure working with you! Hope to see this merged soon

lib/index.js
```diff
@@ -85,7 +85,7 @@ class ReactRefreshPlugin {
     }
 
     // Inject react-refresh context to all Webpack entry points
-    compiler.options.entry = injectRefreshEntry(compiler.options.entry, this.options);
+    compiler.options.entry = injectRefreshEntry(compiler.options.entry, this.options, compiler);
```
pmmmwh, line 88: I'm not quite certain about passing down the whole compiler object here, when what we only need is `compiler.optoins.entry` and `compiler.options.devServer` - can you consolidate the function to only accept `compiler.options` (i.e. `compilerOptions`) and `this.options` (i.e. `options`)?

sockets/utils/getSocketUrlParts.js
```diff
@@ -82,6 +82,15 @@ function getSocketUrlParts(resourceQuery) {
   pathname = parsedQuery.sockPath || pathname;
   port = parsedQuery.sockPort || port;
 
+  if (!hostname || !pathname || !port) {
+    throw new Error(
+      '[React Refresh] Failed to get an URL for the socket connection. ' +
```
pmmmwh, line 87: Me personally prefer using `Array.join` :)

sockets/utils/getSocketUrlParts.js
```diff
@@ -82,6 +82,15 @@ function getSocketUrlParts(resourceQuery) {
...
+        'This might occur when the execution script doesn’t have `src` attribute set. ' +
+        'Either specify it or use overlay options ' +
```
pmmmwh, line 89: This seems a bit difficult to action to me. What exactly are we expecting the user to do?

test/unit/injectRefreshEntry.test.js
```diff
@@ -138,6 +138,61 @@ describe('injectRefreshEntry', () => {
     ]);
   });
 
+  it('should append resource queries to the overlay entry when compiler is specified', () => {
```
pmmmwh, line 141:
```suggestion
  it('should append resource queries to the overlay entry when specified in devServer options', () => {
```

test/unit/injectRefreshEntry.test.js
```diff
@@ -138,6 +138,61 @@ describe('injectRefreshEntry', () => {
...
+        }
+      )
+    ).toStrictEqual([
+      ReactRefreshEntry,
+      `${ErrorOverlayEntry}?sockHost=localhost&sockPath=/socket&sockPort=9000`,
+      'test.js',
+    ]);
+  });
+
+  it('should append resource queries to the overlay entry when both overlay options and compiler are specified', () => {
```
pmmmwh, line 167:
```suggestion
  it('should append resource queries to the overlay entry when specified in both devServer options and overlay options', () => {
```

pmmmwh:
Had a busy week - finally got some free time to review this. I think this is almost good to go!

lib/index.js
frehner: I can do that, but I think it will actually be quite a lot of work - `injectRefreshEntry` recursively calls itself in a lot of different places, and if we make this change then it would require all those recursive calls to to be wrapped in an object

for example, line 119 in `injectRefreshEntry.js` is 

```js
        injectRefreshEntry(resolvedEntry, options)
```

and we would probably have to change it to 

```js
        injectRefreshEntry({entry: resolvedEntry}, options)
```

in addition to updating all the tests in `injectRefreshEntry.test.js`. Again, I don't mind doing that, I just want to be sure that is what want to happen.

Alternatively, I could just change the third parameter to just taking in the `devServer` object and we wouldn't have to churn all the code/tests.

Let me know what you prefer.

sockets/utils/getSocketUrlParts.js
frehner: good point.

Maybe we could say something like 

`You can specify sockPort, sockPath, and sockHost in either webpack-dev-server's config, or in react-refresh-webpack-plugin's config` or something like that?

sockets/utils/getSocketUrlParts.js
```diff
@@ -82,6 +82,15 @@ function getSocketUrlParts(resourceQuery) {
...
+        'https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/main/docs/API.md#overlay'
```
pmmmwh, line 90: How does this sound?

```suggestion
    throw new Error([
      '[React Refresh] Failed to get an URL for the socket connection.',
      'This usually means that the current executed script doesn't have a `src` attribute set.',
      'You should either specify the socket path parameters under the `devServer` key in your Webpack config, or use the `overlay` option.',
      'https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/main/docs/API.md#overlay'
      ].join('\n')
    )
```

lib/index.js
pmmmwh: This is great point. Now that I look at the code, I realised that the options passed down are exclusively used for the overlay entry (and I think for your change you also need to pass the `compiler.options.devServer` object down to the recursion below to ensure that those entry types also get the correct metadata injected). We could probably do something a bit smarter - that is to extract out that part of the logic into some sort of a `getOverlayEntry` function, which does not need to recurse. That way we could keep the recursion as is without having to pollute it, and still ensuring the overlay entries and resource query is being correctly constructed.

What do you think?

frehner: ah, hmm... where are you thinking that `getOverlayEntry` would be called? if we call it inside of `injectRefreshEntry` then we would still have to pass in the compiler stuff to `injectRefreshEntry` in the end, right? 

(I could be missing something though 🙂 )

sockets/utils/getSocketUrlParts.js
frehner: looks good to me!

lib/index.js
pmmmwh: We could call it outside of `injectRefreshEntry` - like above the re-assign of `compiler.options.entry` - then pass the constructed error overlay entries so that theys would only be constructed once?

frehner: ok I think I see where you're going! 😄 

I think I can get an updated version in 10 minutes or so, to see what you think

frehner: ah but that would still require a rework of what is passed into `injectRefreshEntry` though, right? (and thus churn on the tests/recursive calls)

maybe I didn't fully get what you're saying

pmmmwh: > ah but that would still require a rework of what is passed into injectRefreshEntry though, right? (and thus churn on the tests/recursive calls)

Yes it would - mostly for tests - but I think this is a better way forward instead of augmenting the amount of parameters we pass into the recursion (and thus sometimes forgetting them) ...
And they are truly only needed for one pass ...

frehner: Is this sort of what you were thinking? https://github.com/frehner/react-refresh-webpack-plugin/pull/2/files

note that I didn't update the tests yet because I wanted to be sure I was doing it right, before I went and changed all the tests.

pmmmwh: Left some review comments. Super thankful for you two to be driving this effort (and the one for `sockProtocol`)!

frehner:
This PR is ready to be reviewed again now :)

pmmmwh:
> This PR is ready to be reviewed again now :) 

I'll take a look (and hopefully merge it) tomorrow. I'm in the home stretch of fixing tests and will pull that in once it's ready.

frehner:
The good news is that the tests pass now. 😄 

Thanks for merging in

lib/utils/getOverlayEntry.js
```diff
@@ -0,0 +1,57 @@
+const querystring = require('querystring');
+
+/**
+ * Creates an object that contains two entry arrays: the prependEntries and overlayEntries
+ * @param {import('../types').NormalizedPluginOptions} options Configuration options for this plugin.
+ * @param {import('webpack').Compiler.options.devServer | undefined} devServer The webpack devServer config
```
pmmmwh, line 6: JSDoc specific syntax

```suggestion
 * @param {import('webpack').Compiler.options.devServer} [devServer] The webpack devServer config
```

lib/utils/getOverlayEntry.js
```diff
@@ -0,0 +1,57 @@
...
+ * @returns {import('./injectRefreshEntry').AdditionalEntries} An object that contains the Webpack entries for prepending and the overlay feature
```
pmmmwh, line 7: Minor thing - I think we should move the type definition here instead of having it in `injectRefreshEntry` for two reasons:

1. This would prevent recursive imports
2. This would conceptually imply what `injectRefreshEntry` consumes is utterly dependent on this function's implementation, but not the other way around.

lib/utils/getOverlayEntry.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+function getOverlayEntry(options, devServer) {
```
pmmmwh, line 9: You think it would be good to change this into an object (I'm not sure if we would revisit this and add more parameters in the future)?

```suggestion
function getOverlayEntry({ devServer, options }) {
```

lib/utils/getOverlayEntry.js
```diff
@@ -0,0 +1,57 @@
...
+
+  const overlayEntries = [
+    // Legacy WDS SockJS integration
+    options.overlay &&
+      options.overlay.useLegacyWDSSockets &&
+      require.resolve('../../client/LegacyWDSSocketEntry'),
+    // Error overlay runtime
+    options.overlay &&
+      options.overlay.entry &&
+      options.overlay.entry + (queryString && `?${queryString}`),
```
pmmmwh, line 51: This is a nitpick on my old code to make it obvious that this would not cause wrong string interpolations when `queryString` is not a string for whatever reason:

```suggestion
      `${options.overlay.entry}${queryString ? `?${queryString}` : ''}`,
```

lib/utils/getOverlayEntry.js
pmmmwh: Then in `injectRefreshEntry` we could just either import the type and use it, or just do `ReturnType<import('./injectRefreshEntry')>`.

lib/utils/getOverlayEntry.js
pmmmwh, line 9: This is a bit of a minor thing - but since we're generating `additionalEntries` here now, maybe it makes sense to rename this (and the file) to `getAdditionalEntries` or `getInjectedEntries`?

test/unit/getOverlayEntry.test.js
```diff
@@ -0,0 +1,78 @@
...
+const ReactRefreshEntry = require.resolve('../../client/ReactRefreshEntry');
+
+const DEFAULT_OPTIONS = {
+  overlay: {
+    entry: ErrorOverlayEntry,
+  },
+};
+
+describe('getOverlayEntry', () => {
+  it('should append legacy WDS entry when required', () => {
```
pmmmwh, line 13: Can we also add a test for the baseline case (i.e. pass in default comes out as default)?

test/unit/getOverlayEntry.test.js
```diff
@@ -0,0 +1,78 @@
...
+          useLegacyWDSSockets: true,
+        },
+      })
+    ).toStrictEqual({
+      overlayEntries: [require.resolve('../../client/LegacyWDSSocketEntry'), ErrorOverlayEntry],
+      prependEntries: [ReactRefreshEntry],
+    });
+  });
+
+  it('should append resource queries to the overlay entry when specified', () => {
```
pmmmwh, line 27:
```suggestion
  it('should append resource queries to the overlay entry when specified in overlay options', () => {
```

pmmmwh (commented):
Had a final look at this and most comments are fairly minor.

Thanks for driving this and addressing all the comments and changes in a timely manner and sorry for dragging this for so long 🙏🏻

pmmmwh:
Also - I'm going to merge #242 tonight and that would incur some extra changes and merge conflicts that I can handle. Would you mind giving me write access to your branch so I can do that?

frehner:
> Also - I'm going to merge #242 tonight and that would incur some extra changes and merge conflicts that I can handle. Would you mind giving me write access to your branch so I can do that?

Updated the PR with all of your feedback. Thanks!

I believe you have access to my `fix-socket-url` branch in the repo - I have the `Allow Edits by Maintainers` checkbox checked here which says you should be able to push to that. Let me know if I need to do anything else though.

frehner:
it was pointed out to me that the code I had previously would _only_ look at devServer's `sockPort` and `sockHost` config items, but the majority of the time devs don't configure those items. webpack-dev-server itself [falls back to using `port` and `host`](https://github.com/webpack/webpack-dev-server/blob/699404b091541242ad3d5f38f1a0022a83ff09b2/client-src/default/utils/createSocketUrl.js#L77-L79) (they use `hostname` in the linked code but this plugin receives it as `host`) for the socket connection if those `sock-` config items aren't passed, so for better compatibility I updated this PR to do that as well.

frehner:
I merged in and (hopefully) correctly resolved the conflicts from the main branch. 

tests don't run on cirlce - says `Unauthorized`. but they pay locally :)

pmmmwh:
> I merged in and (hopefully) correctly resolved the conflicts from the main branch.

This is so great!!! Thanks so much 🙇🏻‍♂️ 

> tests don't run on cirlce - says `Unauthorized`. but they pass locally :)

I'm trying to get the secrets working for forked builds with Docker Hub authenticated pulls because of their new rate limit 🤦🏻

lib/utils/getAdditionalEntries.js
```diff
@@ -0,0 +1,61 @@
...
+    (sockHost || host) && (resourceQuery.sockHost = sockHost ? sockHost : host);
+    sockPath && (resourceQuery.sockPath = sockPath);
+    (sockPort || port) && (resourceQuery.sockPort = sockPort ? sockPort : port);
+  }
+
+  if (options.overlay) {
+    options.overlay.sockHost && (resourceQuery.sockHost = options.overlay.sockHost);
+    options.overlay.sockPath && (resourceQuery.sockPath = options.overlay.sockPath);
+    options.overlay.sockPort && (resourceQuery.sockPort = options.overlay.sockPort);
+    options.overlay.sockProtocol && (resourceQuery.sockProtocol = options.overlay.sockProtocol);
```
pmmmwh, line 28: We can actually also derive the value of this from the `https` option in `devServer`

pmmmwh (commented):
One more thing 👀

lib/utils/getAdditionalEntries.js
frehner: so up above in the `if(devServer)` block, I should add a line like:

```js
https && (resourceQuery.sockProtocol = 'https');
```

that look right?

pmmmwh:
```js
resourceQuery.sockProtocol = https ? 'https' : 'http';
```

I think dev server defaults to http?

frehner: oh hmm it says that it accepts an object for https as well, but doesn't give an example of what that object would look like

any ideas?

https://webpack.js.org/configuration/dev-server/#devserverhttps

pmmmwh: This should be it - while looking at the docs I also realised that the user could set `http2` which would default to `https` too.

[Ref](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js)

```js
this.options.https = {
  ca: this.options.ca,
  pfx: this.options.pfx,
  key: this.options.key,
  cert: this.options.cert,
  passphrase: this.options.pfxPassphrase,
  requestCert: this.options.requestCert || false,
  /** @deprecated */
  spdy: {
    protocols: ['h2', 'http/1.1'],
  },
};
```

frehner: ok so if it's an object we can just assume we need https then it seems.

so it should be?

```js
resourceQuery.sockProtocol = https || http2 ? 'https' : 'http';
```

pmmmwh: Yup exactly!

frehner:
added that code and 2 additional tests for it

pmmmwh (approved):
LGTM - will merge tonight!

frehner:
@pmmmwh just wondering when you plan on creating a new release for the plugin so this change can go out?

thanks again for your work on this 🙂

pmmmwh:
> @pmmmwh just wondering when you plan on creating a new release for the plugin so this change can go out?

I'm working to have `0.5.0` out this week (yes, a minor 😉)  - which obviously would include this.

PR #242, opened 2020-11-01
feat: add sockProtocol option to overlay

patrikholcak:
Allow forcing protocol for socket. This fixes error in non-standard environments like running the page with "data:" protocol. In that case it is impossible to know if the communication requires http or https.

All in all, having more configuration where it makes sense is always good and having a way to configure the protocol is one thing that seems to always be overlooked (eg. create-react-app)


docs/API.md
```diff
@@ -185,6 +186,17 @@ Type: `string`
 Set a custom path for the error overlay to listen to Webpack build messages.
 Useful if you set `devServer.sockPath` to something other than `/sockjs-node`.
 
+### `sockProtocol`
+
+Default: parsed from current url
+
+Type: `enum`
```
pmmmwh, line 193: I think it would be better for use to be explicit about this - what about:

```suggestion
Type: `http:` or `https:`
```

(refernce `sockIntegration`)

pmmmwh: As a note to that, I think `ws` and `wss` are also technically valid choices if people use strict, native WebSockets (not `SockJS`)

docs/API.md
```diff
@@ -185,6 +186,17 @@ Type: `string`
...
+
+**This is relevant for `webpack-dev-server` only.**
+
+Force protocol type for the error overlay to listen to Webpack build messages.
+Useful if you want to force https communication.
```
pmmmwh, line 198:
```suggestion
Force a protocol for the error overlay to listen for Webpack build messages.
Useful if you want to enforce https communication, or if you're working under a non-HTTP path.
```

types/types.d.ts
```diff
@@ -23,6 +23,10 @@ export type ErrorOverlayOptions = {
    * The socket port to use (WDS only).
    */
   sockPort?: number;
+  /**
+   * The socket protocol to use (WDS only).
+   */
+  sockProtocol?: number;
```
pmmmwh, line 29: Hmm I don't think this is right? Can you run the `generate-types` command to regenerate these files?

docs/API.md
pmmmwh: Another note I think would be relavent - I'm more inclined to have `http` instead of `http:` here. I think we should either go the full way (`http://`) or just go with the protocol name (`http`).

patrikholcak: Yup, I wanted to do that, just wasn’t sure. Updated

types/types.d.ts
patrikholcak: The problem was I didn’t run the command and did it manually. Good catch. Ran the command and it seems to be ok now!

test/unit/validateOptions.test.js
```diff
@@ -381,6 +381,34 @@ describe('validateOptions', () => {
...
+
+  it('should accept "overlay.sockProtocol" when it is "https"', () => {
+    expect(() => {
+      new ReactRefreshPlugin({
+        overlay: { sockProtocol: 'https' },
+      });
+    }).not.toThrow();
+  });
+
+  it('should reject "overlay.sockProtocol" when it is not "http" or  "https"', () => {
```
pmmmwh, line 400: Minor nitpick ;)

```suggestion
  it('should reject "overlay.sockProtocol" when it is not "http" nor  "https"', () => {
```

docs/API.md
```diff
@@ -82,6 +82,7 @@ interface ErrorOverlayOptions {
   sockHost?: string;
   sockPath?: string;
   sockPort?: number;
+  sockProtocol?: 'http' | 'https';
```
pmmmwh, line 85: I did a quick search on GitHub and I think `ws` and `wss` protocols are getting more traction - we should also allow them here (please update all other places related to this enum too!)

sockets/utils/getSocketUrlParts.js
```diff
@@ -82,6 +82,11 @@ function getSocketUrlParts(resourceQuery) {
   pathname = parsedQuery.sockPath || pathname;
   port = parsedQuery.sockPort || port;
 
+  // Make sure the protocol from resource query has a trailing colon
+  if (parsedQuery.sockProtocol) {
+    protocol = url.format({ protocol: parsedQuery.sockProtocol });
```
pmmmwh, line 87: Hmm. Should we just manually format the protocol ourselves?

Technically speaking it is safe to just to add a `:`, since we any protocol received from the Webpack options that does not conform the the `enum` will throw?

docs/API.md
```diff
@@ -190,7 +190,7 @@ Useful if you set `devServer.sockPath` to something other than `/sockjs-node`.
 
 Default: parsed from current url
 
-Type: `http` or `https`
+Type: `http`, `https`, `ws`, `wss`
```
pmmmwh, line 193:
```suggestion
Type: `http`, `https`, `ws` or `wss`
```

pmmmwh (approved):
One last nitpick in my preference of grammar. Otherwise good to go.

However this might have to wait a bit for me to fix the flimsy chromium tests first before merge to make sure stuff pass 🙈

patrikholcak:
@pmmmwh Let me know if you need any help with the tests

PR #273, opened 2020-12-13
feat: implement monorepo compatibility layer with library

### LINKED ISSUE #139: Two Reacts with fast refresh, one overlay won't dismiss
I have two React's with fast refresh, one coming from webpack/react-refresh-webpack-plugin, and the other coming from next.js.  (I think that's purely an implementation detail, though, this bug looks like it would occur with two webpacks as well.)

When I create a runtime error in either project, both error overlays throw.

When I fix my error, the codebase with the error removes their respective overlay, but the other overlay remains.

I think I see what's going on.  Both are catching the runtime error with `window.addEventListener('error')`, but only try to remove their overlay when a reload comes from their codebase.

Is there any way to only catch errors associated with the webpack build?  One idea is to filter errors on the "filename" in the error event, and only throw the overlay if the error is coming from a file in the webpack output:
https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent

### LINKED ISSUE #239: Feature Request: safeThis.__reactRefreshInjected respects webpack library name
Under micro frontend architecture, there will be multiple webpack instances run simultaneously in the same client. Like webpack doing, it appends `output.library` in config to some global variables, like `webpackHotUpdate`, `webpackJsonp`. 

For Example, for project Demo while setting `output.library` to `demo`, `webpackHotUpdate` becomes `webpackHotUpdatedemo`, `webpackJsonp` becomes ``webpackJsonpdemo`.

So, is it possible that this plugin also make the same rule for `__reactRefreshInjected`? Otherwise, it will not work with such case mentioned above.

https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/main/client/ReactRefreshEntry.js#L5

### LINKED ISSUE #250: ReactRefresh in monorepo fails to update when more than one react-app present in the page
Unfortunately it is time consuming to provide a repro repository for this but if nothing pops into mind as a fix i will do it.

in a typescript monorepo (lerna and yarn workspaces enabled) with 2 cra started simultaneous in dev mode, DOM updates do not happen as expected.

i have a index.html that loads code from both react apps (meaning it has two different root divs) so i have 2 bundles present and 2 ws monitoring updates from react-refresh.

crashes in any of the apps get the overlay to show up correctly. the apps load correctly and everything compiles correctly.

the problem is, changes do not get reflected in the browser. if i remove one of the bundles being loaded, everything is back working.

yes, everything is freshly installed. using yarn, removed node_modules and yarn lock, in root and in packages.

## Environment Info:

current version of create-react-app: 4.0.0
running from /Users/.../.config/yarn/global/node_modules/create-react-app

System:
OS: macOS 10.15.7
CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
Binaries:
Node: 10.15.3 - /usr/local/opt/node@10/bin/node
Yarn: 1.22.0 - /usr/local/bin/yarn
npm: 6.4.1 - /usr/local/opt/node@10/bin/npm
Browsers:
Chrome: 86.0.4240.183
Edge: 86.0.622.63
Firefox: Not Found
Safari: 14.0
npmPackages:
react: 17.0.1
react-dom: 17.0.1
react-scripts: 4.0.0
npmGlobalPackages:
create-react-app: Not Found

## Steps to reproduce

    create monorepo
    have index.html load 2 react-apps
    start with lerna run start
    change one file and wait for updates

pmmmwh:
This PR implements a compatibility layer to allow multiple instances of React/React-Refresh to run simultaneously. Another layer of compatibility is added to the overlay to make sure only the first injection would happen (since it is global).

~**TODO: Documentation**~

Fixes #139
Fixes #239
Fixes #250

### DIFF
--- a/client/ErrorOverlayEntry.js
+++ b/client/ErrorOverlayEntry.js
@@ -70,17 +70,22 @@ function compileMessageHandler(message) {
   }
 }
 
-// Only register if we're in non-production mode and if window is defined
 if (process.env.NODE_ENV !== 'production' && typeof window !== 'undefined') {
-  // Registers handlers for compile errors
-  __react_refresh_init_socket__(compileMessageHandler, __resourceQuery);
-  // Registers handlers for runtime errors
-  errorEventHandlers.error(function handleError(error) {
-    hasRuntimeErrors = true;
-    __react_refresh_error_overlay__.handleRuntimeError(error);
-  });
-  errorEventHandlers.unhandledRejection(function handleUnhandledPromiseRejection(error) {
-    hasRuntimeErrors = true;
-    __react_refresh_error_overlay__.handleRuntimeError(error);
-  });
+  // Only register if no other overlay have been registered
+  if (!window.__reactRefreshOverlayInjected) {
+    // Registers handlers for compile errors
+    __react_refresh_init_socket__(compileMessageHandler, __resourceQuery);
+    // Registers handlers for runtime errors
+    errorEventHandlers.error(function handleError(error) {
+      hasRuntimeErrors = true;
+      __react_refresh_error_overlay__.handleRuntimeError(error);
+    });
+    errorEventHandlers.unhandledRejection(function handleUnhandledPromiseRejection(error) {
+      hasRuntimeErrors = true;
+      __react_refresh_error_overlay__.handleRuntimeError(error);
+    });
+
+    // Mark overlay as injected to prevent double-injection
+    window.__reactRefreshOverlayInjected = true;
+  }
 }
--- a/client/ReactRefreshEntry.js
+++ b/client/ReactRefreshEntry.js
@@ -1,13 +1,21 @@
+/* global __react_refresh_library__ */
+
 const safeThis = require('./utils/safeThis');
 
 if (process.env.NODE_ENV !== 'production' && typeof safeThis !== 'undefined') {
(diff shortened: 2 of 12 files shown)


frehner:
@pmmmwh in webpack5, webpack looks at either `output.library`, or `package.json`'s `name` field, or `output.uniqueName`

Here's the docs on that https://webpack.js.org/configuration/output/#outputuniquename

Should we update this to accept those different fields?

pmmmwh:
> @pmmmwh in webpack5, webpack looks at either `output.library`, or `package.json`'s `name` field, or `output.uniqueName`
> 
> 
> 
> Here's the docs on that https://webpack.js.org/configuration/output/#outputuniquename
> 
> 
> 
> Should we update this to accept those different fields? 

I guess the simplest way out is to let Webpack do its thing, look at `output.uniqueName`and if not found default to `output.library`?

frehner:
> > @pmmmwh in webpack5, webpack looks at either `output.library`, or `package.json`'s `name` field, or `output.uniqueName`
> > Here's the docs on that https://webpack.js.org/configuration/output/#outputuniquename
> > Should we update this to accept those different fields?
> 
> I guess the simplest way out is to let Webpack do its thing, look at `output.uniqueName`and if not found default to `output.library`?

I haven't looked too closely at it to know for sure, but that may work.

PR #288, opened 2020-12-27
fix: ipv6 localhost

chenxsan:
There seems to be a bug in the deprecated `url.format()` https://github.com/nodejs/node/issues/36654 when formatting ipv6 url:

```
const url = require('url')
console.log(url.format({
    protocol: 'http',
    hostname: '[::]'
}))
```
We'll get result of `http://[[::]]`. Same behavior applies to `native-url`.

Therefore `new SocketClient()` would run with url `http://[[::]]:8080/sockjs-node` when `hostname` is `[::]`, and that is wrong.

Although they're going to fix it on the Node.js side, but I think it would still make sense to include a quick fix here as it would take some time for `native-url` to update.

### DIFF
--- a/sockets/utils/formatUrl.js
+++ b/sockets/utils/formatUrl.js
@@ -1,0 +1,12 @@
+const url = require('native-url');
+/**
+ * @param {import('./getSocketUrlParts/SocketUrlParts')} urlParts
+ * @returns {string}
+ */
+module.exports = function formatUrl(urlParts) {
+  const myUrl = new URL('http://./');
+  for (const [key, value] of Object.entries(urlParts)) {
+    myUrl[key] = value;
+  }
+  return url.format(myUrl);
+};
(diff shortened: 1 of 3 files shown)


sockets/utils/formatUrl.js
```diff
@@ -0,0 +1,12 @@
+const url = require('native-url');
+/**
+ * @param {import('./getSocketUrlParts').SocketUrlParts} urlParts
+ * @returns {string}
+ */
+module.exports = function formatUrl(urlParts) {
+  const myUrl = new URL('http://./');
+  for (const [key, value] of Object.entries(urlParts)) {
+    myUrl[key] = value;
```
chenxsan, line 9: Since the type of `hostname` is `string`, so it's safe to assign its value to `myUrl.hostname`, otherwise we would get result like `http://null`.

chenxsan:
I've reported the case on node.js https://github.com/nodejs/node/issues/36654

pmmmwh:
Hey, thanks for the PR! We have had our fair share of "bugs" from `native-url` for a while now that I'm considering to actually roll our own URL formatting utility (it is not that difficult given that all the heavy lifting had been done by the `URL` implementation).

I'll leave this open while I decide (and maybe work on that).

Reference route that we could take from Node.js: https://github.com/nodejs/node/pull/28482

PR #291, opened 2021-01-10
fix the sloppy `Uncaught ReferenceError: $RefreshReg$ is not defined` troubleshoot suggestion 

IlanBeloglovsky:
remove unnecessary `$` sign in line 61


pmmmwh (approved):
Thanks

PR #305, opened 2021-02-05
Update TROUBLESHOOTING.md

Slapbox:
Explicitly mention how to handle Electron Main process.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -39,9 +39,9 @@ module.exports = {
 };
 ```
 
-## Usage with Indirection (like Workers and JS Templates)
+## Usage with Indirection (like Workers, JS Templates)
 
-If you share the Babel config for files in an indirect code path (e.g. Web Workers, JS Templates with partial pre-render) and all your other source files, you might experience this error:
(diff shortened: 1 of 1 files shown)


pmmmwh:
Hello, thanks for the PR! Can you explain a bit more what you mean by Electron main process and why does it have this problem?

Slapbox:
Hey @pmmmwh thanks for your reply!

The Electron "main" process is a nodejs process that doesn't have DOM support. The other process type in Electron is the "renderer" process, which displays a window with DOM support.

In our case our Webpack configuration uses the same `babel-loader` block in both the main and renderer processes, which react refresh didn't like since the main process has no DOM.

https://www.electronjs.org/docs/tutorial/quick-start#main-and-renderer-processes

pmmmwh:
I don't think the DOM not existing is an issue? There is no code in the runtime for React Refresh that actually depends on the DOM - I would assume because you did not process the `main` process with this plugin but applied the `react-refresh/babel`? I believe I have stated in the docs that these two have to exist together, is that not the case?

Slapbox:
Oh you know what, you're right on. We were applying the webpack plugin to only the `renderer` process, but we were applying the babel plugin to both `main` and `renderer`, so I think we can close out this PR without merging. Hopefully this issue will help anyone who runs into the same issue as me without the need to clutter up the TROUBLESHOOTING.md file.

Thanks so much for sharing your amazing work @pmmmwh and for taking the time to maintain too!

PR #323, opened 2021-03-02
add default value for protocol

nticaric:
This fixes a bug when the script is called like `<script src="//localhost:8080/js/index.js"></script>`

In this case, the protocol will be `null` and the script will throw an error:

`Uncaught DOMException: Failed to construct 'WebSocket': The URL '//localhost:8080/sockjs-node' is invalid.`


pmmmwh:
If you scroll a bit further down in L70 we actually sets the default to `window.location.protocol`. I think we can just relax that check a bit more to ensure we assign a protocol also when `protocol` itself is empty.

nticaric:
If I'm correctly interpreting the code, then we don't  need lines 66-71 at all if we have at the beginning 
` let protocol = urlParts.protocol || window.location.protocol;`

This would make the code much more readable and cleaner then polluting the if statement with another check if the protocol is null

pmmmwh:
Unfortunately I've been quite busy lately and also a bit focused on other parts of the plugin at the moment. I'll take a look at this during the weekend.

nticaric:
Did you have any chance to review this?

pmmmwh:
Hi, I did some testing and I'm not sure if I see how this will fix the issue. If you revert your changes but keep the tests, they currently already passes.

This line below actually accounts for this behaviour (The bracketing currently is wrong though):

```js
if ((hostname && isEmptyHostname) || window.location.protocol === 'https:')) {}
````

When the script source is `//localhost:8080/js/index.js`, `url.parse` will return something like so:

```js
{
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: '//localhost:8080/js/index.js',
  path: '//localhost:8080/js/index.js',
  href: '//localhost:8080/js/index.js'
}
```

As you can see, both `hostname` and `protocol` are `null`, which fulfills the condition for `hostname` to be re-assigned, and thus fulfilling the condition of `hostname && isEmptyHostname`.

nticaric:
On a real project when the script source is `<script src="//localhost:8080/js/index.js"></script>`  `url.parts` returns this:

```
auth: null
hash: null
host: "localhost:8080"
hostname: "localhost"
href: "//localhost:8080/js/index.js"
path: "/js/index.js"
pathname: "/js/index.js"
port: "8080"
protocol: null
query: null
search: null
slashes: true
```

As you can see, in this case the `hostname` is not null and the if statement won't catch it

nticaric:
Just changed the unit tests, so they will now fail with the old code

pmmmwh:
See #332 - that should fix this. We're trying to move to a direction where we would not depend on `native-url` (or the "deprecated" URL API implementations borrowed from Node.js), and I think that is a good step forward.

PR #331, opened 2021-03-18
fix: fix getCurrentScriptSource to handle scripts without src

tinytinyeye:
The current implementation of `getCurrentScriptSource` will fail when there are scripts without `src` used in the document and the script that runs React app is not the last script in the document.

For example, in the case when there are other 3rd party script snippets such as Qualtrics used in the document, the last script get by the original implementation could be such script and the `src` will be null. As a result, `getCurrentScriptSource` will return `null` and `getSocketUrlParts` will fail at `url.parse(scriptSource || '')`.

To fix this issue, we could return the last script that HAS `src` attribute as the current script.

### DIFF
--- a/sockets/utils/getCurrentScriptSource.js
+++ b/sockets/utils/getCurrentScriptSource.js
@@ -11,7 +11,8 @@ function getCurrentScriptSource() {
 
   // Fallback to getting all scripts running in the document.
   const scriptElements = document.scripts || [];
-  const currentScript = scriptElements[scriptElements.length - 1];
+  const nonNullScriptElements = [...scriptElements].filter((elm) => elm.getAttribute('src'));
+  const currentScript = nonNullScriptElements[nonNullScriptElements.length - 1];
   if (currentScript) {
     return currentScript.getAttribute('src');
   }
--- a/sockets/utils/getCurrentScriptSource.js
+++ b/sockets/utils/getCurrentScriptSource.js
@@ -11,9 +11,11 @@ function getCurrentScriptSource() {
 
   // Fallback to getting all scripts running in the document.
   const scriptElements = document.scripts || [];
-  const nonNullScriptElements = [...scriptElements].filter((elm) => elm.getAttribute('src'));
-  const currentScript = nonNullScriptElements[nonNullScriptElements.length - 1];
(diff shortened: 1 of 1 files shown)


nticaric:
If the changes in https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/323 go through this will be automatically fixed

sockets/utils/getCurrentScriptSource.js
```diff
@@ -11,7 +11,8 @@ function getCurrentScriptSource() {
 
   // Fallback to getting all scripts running in the document.
   const scriptElements = document.scripts || [];
-  const currentScript = scriptElements[scriptElements.length - 1];
+  const nonNullScriptElements = [...scriptElements].filter((elm) => elm.getAttribute('src'));
```
pmmmwh, line 14: Unfortunately, we cannot use ES6 syntax here because we also need to support IE and older browsers.

```js
const scriptElementsWithSrc = Array.prototype.filter.call(scriptElements, function (elem) {
  return elem.getAttribute("src");
});
```

sockets/utils/getCurrentScriptSource.js
```diff
@@ -11,7 +11,8 @@ function getCurrentScriptSource() {
...
+  const currentScript = nonNullScriptElements[nonNullScriptElements.length - 1];
   if (currentScript) {
     return currentScript.getAttribute('src');
   }
```
pmmmwh, line 18: I think here we can just check if the length of the filtered elements is non-zero and return the `src`.

sockets/utils/getCurrentScriptSource.js
tinytinyeye: Thanks for the quick reply! I can update that.

sockets/utils/getCurrentScriptSource.js
tinytinyeye: Yeah, I think it would be good to check if the length is non-zero.

tinytinyeye:
Hey @pmmmwh, recently I updated my commits to include the changes you suggested and was wondering if you still have other concerns regarding my fix. Thank you!

pmmmwh (approved):

PR #351, opened 2021-04-12
small alt to remove the need of the filter function in webpack config example

joaomelo:
1. Thanks for the awesome project! Just plug it in an app, and it works beatifully.
2. I propose that the config would be clearer without the need for the filter function, but I admit that this assumption can be considered personal taste.  Anyway...

The config presented in the README uses the filter function to removed false values from the arrays in case of production builds. The PR proposes alternative expressions that dispense the filter function. 

Refs:
- A working build pipeline [example](https://github.com/joaomelo/arc).
- Some blog [post](https://amberley.dev/blog/2020-09-07-conditionally-add-to-array-or-obj/) diving in the notation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,8 +99,8 @@ module.exports = {
               // ... other options
               plugins: [
                 // ... other plugins
-                isDevelopment && require.resolve('react-refresh/babel'),
-              ].filter(Boolean),
+                ...isDevelopment ? [require.resolve('react-refresh/babel')] : [],
+              ],
             },
           },
         ],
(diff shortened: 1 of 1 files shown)


pmmmwh:
As you have said, this is a stylistic preference.
I prefer using the `.filter` apporach as it doesn't require creation of many intermediate arrays, and I think it is much clearer to people new to setting up these things up from my experience.

Thanks for the PR, but I'll close it for now.
We'll try to rewrite examples in the future so they can be clearer/cleaner and maybe then I'll revisit this.

PR #374, opened 2021-04-26
fix: prevent overlay code from running if disabled

keraf:
Setting the `overlay` property to `false` in the plugin options blows up when loading the page. The following error message is displayed in the latest beta release (0.5.0-beta.4, same issue is present in 0.4.3):
Screenshot from 2021-04-26 15-25-24

Checking if `__react_refresh_socket__` is truth-y does the trick. `__react_refresh_socket__` is set to false when the overlay is disabled (see [lib/index.js#L101](https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/main/lib/index.js#L101)).

### DIFF
--- a/client/ErrorOverlayEntry.js
+++ b/client/ErrorOverlayEntry.js
@@ -74,7 +74,7 @@ function compileMessageHandler(message) {
 if (process.env.NODE_ENV !== 'production' && typeof window !== 'undefined') {
   runWithPatchedUrl(function setupOverlay() {
     // Only register if no other overlay have been registered
-    if (!window.__reactRefreshOverlayInjected) {
+    if (!window.__reactRefreshOverlayInjected && __react_refresh_socket__) {
       // Registers handlers for compile errors
       __react_refresh_socket__.init(compileMessageHandler, __resourceQuery);
       // Registers handlers for runtime errors


pmmmwh:
Hmm, this seems a bit weird to me. 

By default the overlay entry would not be injected if `overlay` is set to false.

Are you using it directly?

(I would happily accept this fix, just wanted to know the root cause of this)

keraf:
I just tried to reproduce it on a different machine and using the webpack-plugin-serve example project. The `ErrorOverlayEntry` client is indeed not injected when `overlay` is set to `false`. But for some reason, in my project it still gets injected, leading to this error (and the app breaking).

No clue why this could be happening. I use a more recent version of `webpack` (5.28.0) and `webpack-plugin-server` (1.4.1). Or it could just be a misconfiguration in my config. I will investigate it tomorrow and keep you updated on my findings.

keraf:
Sorry for the late reply. I've updated the dependencies in the example projects just to make sure and it works as expected without this "fix". I guess I must have fudged something in my configuration. I will keep you updated if I find out why.

pmmmwh:
I'll move forward to merge this in the spirit of keeping things in the plugin more "re-usable" when something custom is needed.

pmmmwh (approved):

PR #480, opened 2021-08-18
docs: make Webpack 5 compat explicit

### LINKED ISSUE #474: The doc should precise it also works for Webpack 5
When I see this table: 
image

It makes me think this plugin doesn't work with Webpack 5 as it's written "Best" for Webpack 4.

Could we adjust the doc to precise it also works with webpack 5?

pmmmwh:
Fixes #474


NMinhNguyen:
Is this for the current stable version or `next` (`0.5.0-rc.5`) only?

PR #484, opened 2021-08-23
chore: update type-fest version

soilSpoon:
npm 7 throws an error when installing packages if the versions of peerDependencies are not compatible.

Since type-fest is a utility package, you will likely use many versions, so you should include as many versions as possible to avoid breaking DX.

Now, this package uses only Except, SetRequired, and LiteralUnion in type-fest, so it is expected to work normally in 0.17.0 or later.

### Reference
https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies
https://github.com/sindresorhus/type-fest/releases/tag/v0.6.0
https://github.com/sindresorhus/type-fest/releases/tag/v0.8.0
https://github.com/sindresorhus/type-fest/releases/tag/v0.17.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "puppeteer": "^9.1.1",
     "react-refresh": "^0.10.0",
     "sourcemap-validator": "^2.1.0",
-    "type-fest": "^2.0.0",
+    "type-fest": "^1.2.1 || ^2.0.0",
     "typescript": "4.3.5",
     "webpack": "^5.42.0",
     "webpack-cli": "^4.7.2",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "@types/webpack": "4.x || 5.x",
     "react-refresh": "^0.10.0",
     "sockjs-client": "^1.4.0",
-    "type-fest": ">=0.17.0 <=2.x",
+    "type-fest": ">=0.17.0 <2.0.0",
     "webpack": ">=4.43.0 <6.0.0",
     "webpack-dev-server": "3.x || >=4.0.0-beta.0",
     "webpack-hot-middleware": "2.x",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -138,9 +138,6 @@
       "optional": true
     }
   },
-  "resolutions": {
```
pmmmwh, line 141: Can we keep this? During development we use it to ensure only one version of `type-fest` is installed.

package.json
```diff
@@ -96,7 +96,7 @@
     "puppeteer": "^9.1.1",
     "react-refresh": "^0.10.0",
     "sourcemap-validator": "^2.1.0",
-    "type-fest": "^1.2.1",
+    "type-fest": "^1.2.1 || ^2.0.0",
```
pmmmwh, line 99: `^2.0.0` doesn't work - particularly because it dropped support for Node 10.x.

In reality we could accept it, but this will fail tests and I don't think it is ensured to work due to Sindre's move to ESM.

```suggestion
    "type-fest": "^1.2.1",
```

package.json
```diff
@@ -112,7 +112,7 @@
     "@types/webpack": "4.x || 5.x",
     "react-refresh": "^0.10.0",
     "sockjs-client": "^1.4.0",
-    "type-fest": "1.x",
+    "type-fest": ">=0.17.0 <=2.x",
```
pmmmwh, line 115:
```suggestion
    "type-fest": ">=0.17.0 <2.0.0",
```

pmmmwh (changes requested):

package.json
soilSpoon: Oh, I see.

pmmmwh (approved):

pmmmwh:
Thanks for your PR, I'll merge once all tests pass.

pmmmwh (approved):

PR #501, opened 2021-09-07
fix: replace ansi-html with ansi-html-community

rap2hpoutre:
This fixes the ReDoS vulnerability [CVE-2021-23424](https://github.com/advisories/GHSA-whgm-jr23-g3j9).

See: https://github.com/Tjatse/ansi-html/issues/19#issuecomment-913119841

### DIFF
--- a/overlay/components/CompileErrorTrace.js
+++ b/overlay/components/CompileErrorTrace.js
@@ -1,4 +1,4 @@
-const ansiHTML = require('ansi-html');
+const ansiHTML = require('ansi-html-community');
 const entities = require('html-entities');
 const theme = require('../theme.js');
 const { formatFilename } = require('../utils.js');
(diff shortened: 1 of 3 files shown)


pmmmwh:
I'll wait a bit on this one.

Given that our use case directly consumes errors from Webpack's stats, the risk of hitting a ReDoS is low. I do notice that `ansi-html` is not really maintained anymore, so we might swap to other alternatives/do something ourselves instead in the long run.

boppaaronPFG (approved):
This is the same fix that was implemented for [webpack-dev-server](https://github.com/webpack/webpack-dev-server/releases/tag/v4.1.1).

rap2hpoutre:
@pmmmwh Thank you for your answer. Since webpack-dev-server have patched it with the same community library, I guess we could try to rely on it too: https://github.com/webpack/webpack-dev-server/releases/tag/v4.1.1. It could be considered a temporary fix though.

pmmmwh (approved):

PR #507, opened 2021-09-14
Bump type-fest in peerDependencies

soilSpoon:
https://github.com/sindresorhus/type-fest/releases/tag/v2.0.0

type-fest is 2.0 and there is no problem with compatibility, so bump

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     "@types/webpack": "4.x || 5.x",
     "react-refresh": "^0.10.0",
     "sockjs-client": "^1.4.0",
-    "type-fest": ">=0.17.0 <2.0.0",
+    "type-fest": ">=0.17.0 <3.0.0",
     "webpack": ">=4.43.0 <6.0.0",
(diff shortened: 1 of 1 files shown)


pmmmwh:
No, `type-fest` v2 does not support Node.js v10 but we support that. I'm not too sure what the proper solution should be, technically it works but it will crash on install due to `engine` being incompatible.

soilSpoon:
> No, `type-fest` v2 does not support Node.js v10 but we support that. I'm not too sure what the proper solution should be, technically it works but it will crash on install due to `engine` being incompatible.

Since you specify it as a range, can't it be 2.x and also 1.x?
I think it's the user's choice.

pmmmwh:
OK, after sitting on this a bit I think it's fair to allow `type-fest` v2, but after merging this I'll also add a note in the README warning that v2 only supports node.js 12+.

pmmmwh (approved):

PR #516, opened 2021-09-28
Support module federation (close #394)

### LINKED ISSUE #394: Component modifications are not applied in module federation
I'm having an issue getting react refresh working on a project with module fedration. The ws connection is working, also in the console it seems its working, but the modifications are not applied to the component. 

```
[webpack-dev-server] App updated. Recompiling...
index.js:3210 [webpack-dev-server] App hot update...
log.js:24 [HMR] Checking for updates on the server...
log.js:24 [HMR] Updated modules:
log.js:24 [HMR]  - ./src/Button.js
log.js:24 [HMR] App is up to date.
```

I have created an basic example: https://github.com/kabel2/module-federation-examples/tree/feature/example-react-refresh/basic-host-remote

Changing the innertext of the button in the "Button.js" component in "app2" does not apply the changes live, only after reload. 

With the old react-hot-loader it seems to work.

ije:
close #394

```
// app1
import * as refresh_runtime from 'react-refresh/runtime';
import ReactDOM from 'react-dom';

Object.assign(window, { __sharing_react_refresh_runtime__: refresh_runtime });
```

then in `app2` the **\_\_sharing_react_refresh_runtime\_\_** will be used

### DIFF
--- a/lib/runtime/RefreshUtils.js
+++ b/lib/runtime/RefreshUtils.js
@@ -1,5 +1,5 @@
 /* global __webpack_require__ */
-var Refresh = require('react-refresh/runtime');
+var Refresh = window.__sharing_react_refresh_runtime__ || require('react-refresh/runtime');
 
 /**
  * Extracts exports from a webpack module object.
--- a/loader/index.js
+++ b/loader/index.js
@@ -46,11 +46,11 @@ function ReactRefreshLoader(source, inputSourceMap, meta) {
 
   const RefreshSetupRuntimes = {
     cjs: Template.asString(
-      `__webpack_require__.$Refresh$.runtime = require('${RefreshRuntimePath}');`
+      `__webpack_require__.$Refresh$.runtime = window.__sharing_react_refresh_runtime__ || require('${RefreshRuntimePath}');`
     ),
     esm: Template.asString([
       `import * as __react_refresh_runtime__ from '${RefreshRuntimePath}';`,
-      `__webpack_require__.$Refresh$.runtime = __react_refresh_runtime__;`,
+      `__webpack_require__.$Refresh$.runtime = window.__sharing_react_refresh_runtime__ || __react_refresh_runtime__;`,
     ]),
   };
 


pmmmwh:
Can't you just mark `react-refresh/runtime` as shared?

ije:
no, it doesn't work if you put it in the `shared`

pmmmwh:
I do not understand the issue and the solution, it seems quite obvious here that the issue is we need a singleton for `react-refresh/runtime`, and by using `window.something` we're effectively enforcing that.

Why wouldn't it work if we put it as a shared package and go through the hoops of setting up externalising React properly?

Also - this package aims to be as portable as possible and not use `window` since that is Web-only, so this solution is probably not going to be an acceptable one.

ije:
yes this is kind of workaround not good. I tried `singleton` but failed. my problem is that i want to develop some sub modules in an online app by dynamic import, and those modules may run in my local device. everything is ok, including the css hmr, only the fresh plugin not work (ws connection is ok, update message accepted but changes not applied). i finally found out this can work for me, anyway i also think this is not a good idea, I am going to close it. thanks

wlshopee:
Hi, I took a look at the issue that this MR was trying to solve, and I researched about marking the `react-refresh/runtime` as shared and singleton through module federation.

Marking it as shared and singleton does not work, because the react-refresh code will be bundled with the project's federated module (and loaded through the federeted entry point like remoteEntry.js). However, the runtime injected by this plugin is executed on the project's main entry point. Since the project's main entry point is always loaded before any federated modules are loaded, trying to import `react-refresh/runtime` will result in this error:

Screenshot 2021-12-07 at 12 28 03

Btw, the error message seems to imply that setting `eagar: true` will solve this error, but it doesn't, because it only changes the location of the code inside the federated module, it doesn't change the loading order of the project's main entry point and the federated modules.

I haven't found any effective alternatives of externalising this library as a singleton using existing Webpack mechanisms.

I understand that the main concern of this MR is that it is kind of a workaround implementation (the window variable doesn't work for non-Web situations). If I tweak the original MR to provide an escape hatch for the runtime library instead, would that be a more acceptable alternative design? @pmmmwh

A rough user example of how the escape hatch might look like:

```js
module.exports = {
    /* ... */
    plugins: [isDevelopment && new ReactRefreshWebpackPlugin({
        refreshLibrary: () => {
           // an escape hatch, return either the loaded library,
           // or 'undefined' to use the normal import
           return window.__sharing_react_refresh_runtime__
        }
    })].filter(Boolean)
};
`

**Comment by pmmmwh** [comment]
Hi, I am really not very knowledgeable on module federation setups, is it possible for you to create a reproducible example so we can work out what is best for the situation? Does it not work to exclude react refresh in all shared modules and make them tap into the runtime on the main entry point? What are the difficulties of externalising the runtime and how is React loaded in this case - can it be done the same way for the runtime?

**Comment by wangleng-shopee** [comment]
> Hi, I am really not very knowledgeable on module federation setups, is it possible for you to create a reproducible example so we can work out what is best for the situation?

Sure, no problem. I have created a minimal repo here: https://github.com/wangleng-shopee/refresh-mod-fed. Run `yarn start` at the root directory, then visit http://localhost:3000. You will see the same error that I mentioned previously. This is when we try to use the singleton mechanism provided by module federation for react refresh.

> What are the difficulties of externalising the runtime and how is React loaded in this case - can it be done the same way for the runtime?

You can see this part of the example repo, I used the exact same singleton mechanism as React:

https://github.com/wangleng-shopee/refresh-mod-fed/blob/8294478b6f3b9d9764da7e89d8fb5483b74e5775/host/webpack.config.js#L37-L40

So why doesn't React produce the same error as react-refresh? That's because we are able to delay the loading of the React library, by doing a dynamic import of the real entry point. You can see that our `index.js` doesn't house the code for our ReactDOM setup, it does a `import("./bootstrap")` instead:

https://github.com/wangleng-shopee/refresh-mod-fed/blob/8294478b6f3b9d9764da7e89d8fb5483b74e5775/host/src/index.js#L1

Without this trick, we will also encounter the same "Shared module is not available for..." error for React. 

We can't play a similar trick for react-refresh, because we don't have control over the plugin's output to do that. 

> Does it not work to exclude react refresh in all shared modules and make them tap into the runtime on the main entry point?

I haven't found any other mechanisms provided by Webpack that allow us to do that. The closest was using `externals` to force webpack to not to even bundle react-refresh's runtime anywhere, but `externals` has two assumptions: (1) you are importing a library and not some local file, (2) the import is always the same name (i.e. `require("react-refresh/runtime")`), so imports like `./node_modules/react-refresh/runtime.js` can't be externalised using this method.

**Comment by danieloprado** [comment]
I agree that is not a final solutions but this is a good workaround.

I just changed to make runtime/RefreshUtils define the window's var, so you don't need to define it at the app1/root.
If anyone want to use this, I recommend to use [patch-package](https://www.npmjs.com/package/patch-package):

filename: patches/@pmmmwh+react-refresh-webpack-plugin+0.5.7.patch
```diff
diff --git a/node_modules/@pmmmwh/react-refresh-webpack-plugin/lib/runtime/RefreshUtils.js b/node_modules/@pmmmwh/react-refresh-webpack-plugin/lib/runtime/RefreshUtils.js
index 9880757..a9c660a 100644
--- a/node_modules/@pmmmwh/react-refresh-webpack-plugin/lib/runtime/RefreshUtils.js
+++ b/node_modules/@pmmmwh/react-refresh-webpack-plugin/lib/runtime/RefreshUtils.js
@@ -1,5 +1,9 @@
 /* global __webpack_require__ */
-var Refresh = require('react-refresh/runtime');
+var Refresh = window.__sharing_react_refresh_runtime__ || require('react-refresh/runtime');
+
+if(!window.__sharing_react_refresh_runtime__) {
+  Object.assign(window, { __sharing_react_refresh_runtime__: Refresh });
+}
 
 /**
  * Extracts exports from a webpack module object.
diff --git a/node_modules/@pmmmwh/react-refresh-webpack-plugin/loader/index.js b/node_modules/@pmmmwh/react-refresh-webpack-plugin/loader/index.js
index e443330..b8f378d 100644
--- a/node_modules/@pmmmwh/react-refresh-webpack-plugin/loader/index.js
+++ b/node_modules/@pmmmwh/react-refresh-webpack-plugin/loader/index.js
@@ -46,11 +46,11 @@ function ReactRefreshLoader(source, inputSourceMap, meta) {
 
   const RefreshSetupRuntimes = {
     cjs: Template.asString(
-      `__webpack_require__.$Refresh$.runtime = require('${RefreshRuntimePath}');`
+      `__webpack_require__.$Refresh$.runtime = window.__sharing_react_refresh_runtime__ || require('${RefreshRuntimePath}');`
     ),
     esm: Template.asString([
       `import * as __react_refresh_runtime__ from '${RefreshRuntimePath}';`,
-      `__webpack_require__.$Refresh$.runtime = __react_refresh_runtime__;`,
+      `__webpack_require__.$Refresh$.runtime = window.__sharing_react_refresh_runtime__ || __react_refresh_runtime__;`,
     ]),
   };
```

PR #531, opened 2021-11-08
fix: always use the regular function to avoid losing module context

latin-1:
Using an arrow function here would cause the loss of context (`this`) inside the module factory, which may break some modules that rely on it.


pmmmwh:
Hey, thanks for the PR. I'll review this tomorrow and also investigate a bit into the issue.

Is there concrete example demonstrating how arrow functions would break?

latin-1:
Assume we have a module, x.js:

```js
this.foo = "bar";
```

And now, we are trying to require it from node:

```sh
$ echo 'this.foo = "bar";' > x.js
$ node -e 'require("./x").foo' -p
bar
```

Webpack has a similar approach to Node.js by calling the module factory with `this = module.exports`. However, the react-refresh-webpack-plugin's runtime code breaks this treatment.

Currently, we are running into this bug with `@mediapipe/face_mesh`, which uses the same way to export something in both browsers and webpack.

lib/utils/makeRefreshRuntimeModule.js
```diff
@@ -27,45 +27,44 @@ function makeRefreshRuntimeModule(webpack) {
...
-            `options.factory = ${runtimeTemplate.basicFunction(
-              'moduleObject, moduleExports, webpackRequire',
-              [
-                `${refreshGlobal}.setup(options.id);`,
-                'try {',
-                webpack.Template.indent(
-                  'originalFactory.call(this, moduleObject, moduleExports, webpackRequire);'
-                ),
-                '} finally {',
+            `options.factory = function(moduleObject, moduleExports, webpackRequire) {`,
```
pmmmwh, line 30:
```suggestion
            `options.factory = function (moduleObject, moduleExports, webpackRequire) {`,
```

lib/utils/makeRefreshRuntimeModule.js
```diff
@@ -27,45 +27,44 @@ function makeRefreshRuntimeModule(webpack) {
...
                 ]),
+                '} else {',
+                webpack.Template.indent(`${refreshGlobal}.cleanup(options.id)`),
                 '}',
-              ]
-            )}`,
+              ]),
+              '}',
+            ]),
+            `}`,
```
pmmmwh, line 67:
```suggestion
            `};`,
```

test/unit/makeRefreshRuntimeModule.test.js
```diff
@@ -119,7 +119,7 @@ __webpack_require__.$Refresh$ = {
     expect(runtime).toMatchInlineSnapshot(`
 "__webpack_require__.i.push((options) => {
 	const originalFactory = options.factory;
-	options.factory = (moduleObject, moduleExports, webpackRequire) => {
+	options.factory = function(moduleObject, moduleExports, webpackRequire) {
```
pmmmwh, line 122:
```suggestion
	options.factory = function (moduleObject, moduleExports, webpackRequire) {
```

pmmmwh (approved):

PR #545, opened 2021-12-18
feat: exclude files that referenced as assets

latin-1:
Sometimes, we may want to include some js files as-is by writing some code like:

```
loadScript(new URL('./vendor.js', import.meta.url).toString())
```

In this case, injecting runtime code cause that file to be broken. We should exclude all asset modules from being injected.

### DIFF
--- a/lib/utils/injectRefreshLoader.js
+++ b/lib/utils/injectRefreshLoader.js
@@ -26,7 +26,7 @@ function injectRefreshLoader(moduleData, injectOptions) {
   if (
     // Include and exclude user-specified files
     match(moduleData.resource) &&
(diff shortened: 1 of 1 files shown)


lib/utils/injectRefreshLoader.js
```diff
@@ -26,6 +26,8 @@ function injectRefreshLoader(moduleData, injectOptions) {
   if (
     // Include and exclude user-specified files
     match(moduleData.resource) &&
+    // Exclude files that referenced as assets
```
pmmmwh, line 29:
```suggestion
    // Exclude files referenced as assets
```

pmmmwh (approved):

PR #641, opened 2022-05-28
fix: support the 24-bit color of ansi escape sequences

lightyen:
Fix:
Screenshot 2022-05-28 111530

To:
Screenshot 2022-05-28 111651


pmmmwh:
I'm not sure if we want this - WDS also does not support 24-bit sequences AFAIK, and tbh I would really want to move away from more ANSI parsing on the error overlay if possible ...

PR #662, opened 2022-07-25
feat: add support for ts-loader without transpileOnly

julienv3:
This change adds support for ts-loader integration without having to set it to transpileOnly.

This is useful since transpileOnly results in increased bundle sizes due to several optimizations becoming impossible, e.g. const enum definition removal.

### DIFF
--- a/lib/utils/injectRefreshLoader.js
+++ b/lib/utils/injectRefreshLoader.js
@@ -35,6 +35,8 @@ function injectRefreshLoader(moduleData, injectOptions) {
     !moduleData.resource.includes(path.join(__dirname, '../runtime/RefreshUtils')) &&
     // Check to prevent double injection
     !moduleData.loaders.find(({ loader }) => loader === resolvedLoader)
+    // Check to prevent interfering with ts-loader watch-run's attempt to apply loaders specified after itself
+    moduleData.loaders.findIndex(({ loader }) => loader.endsWith("stringify-loader.js")) !== 0
   ) {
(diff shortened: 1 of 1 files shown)


lib/utils/injectRefreshLoader.js
```diff
@@ -35,6 +35,8 @@ function injectRefreshLoader(moduleData, injectOptions) {
     !moduleData.resource.includes(path.join(__dirname, '../runtime/RefreshUtils')) &&
     // Check to prevent double injection
     !moduleData.loaders.find(({ loader }) => loader === resolvedLoader)
+    // Check to prevent interfering with ts-loader watch-run's attempt to apply loaders specified after itself
+    moduleData.loaders.findIndex(({ loader }) => loader.endsWith("stringify-loader.js")) !== 0
```
julienv3, line 39: This is the only problem I've had with ts-loader integration without transpileOnly.

The loader was executing as part of the code in 
https://github.com/TypeStrong/ts-loader/blob/12f4ffe738abcdb2ad7f18d6f6724d89d8cc1cfc/src/watch-run.ts
that attempts to run loaders specified after ts-loader (refresh being before it so shouldn't run) with stringify-loader being the last one, and then watch-run attempts to JSON.parse the result.

The problem is this injection logic added its loader before stringify-loader, which resulted in JSON.parse crashing on parsing the output of the loadModule call since this loader adds code around its input (in this case stringified code).

pmmmwh:
Seemingly no-longer necessary.

Issue #543, opened 2021-12-16
swc-loader much slower if react refresh plugin is enabled


piotrpalek:
Hey, I've tried integrating react fast refresh with `swc-loader`, and the result is that if I add `new ReactRefreshPlugin()` to the `plugins: []` list in webpack, `swc-loader` is _dog slow_ (well, as slow as if I was using babel). If I remove this plugin it's way faster, any ideas what could be wrong? Is this expected?

pmmmwh:
> Hey, I've tried integrating react fast refresh with `swc-loader`, and the result is that if I add `new ReactRefreshPlugin()` to the `plugins: []` list in webpack, `swc-loader` is _dog slow_ (well, as slow as if I was using babel). If I remove this plugin it's way faster, any ideas what could be wrong? Is this expected?

Can you share a bit more about your setup, or create a reproducible example? There have been performance fixes in the `0.5.x` releases, but we do inject another loader into the bundling chain so maybe that would slow things down.

piotrpalek:
Hey! Thanks for the info :)

Sorry if I'm light on details, but I'm still trying to come up with a reproducible example, and I can't share the whole repo since it's work related. I was asking to see if maybe this was known, could help me in research knowing where to look :)

I'll be hacking around and let you know if/when I manage to reproduce it or solve the issue

pmmmwh:
> Thanks for the info :)
> 
> Sorry if I'm light on details, but I'm still trying to come up with a reproducible example, and I can't share the whole repo since it's work related. I was asking to see if maybe this was known, could help me in research knowing where to look :)
> 
> I'll be hacking around and let you know if/when I manage to reproduce it or solve the issue

It could very much be due to source map generation, or if you have a lot of files that needs to be processed. Those are the two most likely points of clogging.

marco2216:
Hi @piotrpalek did you find any solution? We are experiencing the same issue on a large project

piotrpalek:
@marco2216 no sorry :( for now we just didn't move to `swc-loader`

marco2216:
@piotrpalek Feel free to try out if this change helps, I'm not actually sure if it has anything to do with swc-loader, but it significantly improved our transpile speed https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/669

PR #669, opened 2022-09-01
Fix injectRefreshLoader performance issue

marco2216:
After investigating some performance issues that only appear when enabling this plugin, I discovered that the `require.resolve('react-refresh')` which gets called in `injectRefreshLoader` seems to be the cause.
Moving the call outside the `injectRefreshLoader` cuts initial compile time in more than half (~40s -> ~15s) and incremental compile time has a similar speedup (~8s -> ~3s).

I don't know why, but I guess the require.resolve is not cached for some reason, or just takes some time, which adds up when the function runs for thousands of modules?

The change moving the `refreshUtilsPath` outside the function is not required to improve the performance, but I don't see why it wouldn't be moved outside the function as it should be static.

Tested on two different machines running the same project. May or may not address this issue: https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/543

### DIFF
--- a/lib/utils/injectRefreshLoader.js
+++ b/lib/utils/injectRefreshLoader.js
@@ -13,6 +13,8 @@ const path = require('path');
  */
 
 const resolvedLoader = require.resolve('../../loader');
+const reactRefreshPath = path.dirname(require.resolve('react-refresh'))
+const refreshUtilsPath = path.join(__dirname, '../runtime/RefreshUtils')
 
 /**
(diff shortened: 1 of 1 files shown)


marco2216:
@pmmmwh Please let me know if you need any additional information to get this merged!

pmmmwh (approved):

pmmmwh:
Thanks!

JSerFeng:
Any idea on why `require.resolve` is not cached 🤔

pmmmwh:
Released.

Issue #245, opened 2020-11-02
TypeError: Cannot read property 'exports' of undefined function getModuleExports(moduleId) {   return __webpack_require__.c[moduleId].exports; }


liuhaidi1209:
My app encountered the following error when I tried integrating @pmmmwh/react-refresh-webpack-plugin.
"@pmmmwh/react-refresh-webpack-plugin": "0.4.2",

![image](https://user-images.githubusercontent.com/16359954/97874554-1e6f4880-1d54-11eb-85cf-9d969db57e4b.png)

https://user-images.githubusercontent.com/16359954/97874554-1e6f4880-1d54-11eb-85cf-9d969db57e4b.png

HOUCe:
Anyone could give this issue a little love plz? Got the same problem here when I upgraded Create-react-app to 4.0.0

pmmmwh:
Hi. For me to be able to look into issues, I would need more information about the actual setup. Things like Webpack version, how your app is being structured, etc. If possible a reproducible example would help with triaging a lot.

The error itself to me seems like the Webpack `module.cache` is not defined - so either it was not provided by Webpack to the module (because we did not declare it - I think explicit declaration for stuff like this is required in Webpack 5), or your module cache is corrupt.

pmmmwh:
Hi, I did a bit of digging and I think if you have **ANY** variable named `module` within your project it will crash Webpack.

scristobal:
it solved my problem and made my day, thank you @pmmmwh !!

iMoses-Apiiro:
@pmmmwh seems that this bug is still happening, so why is this issue closed?

pmmmwh:
> @pmmmwh seems that this bug is still happening, so why is this issue closed?

Do you have a variable named module? That is working as intended, reusing special Webpack injected variables will break things as expected.

iMoses-Apiiro:
You're talking about a global `module` variable as part of the commonjs standard, or any local variable named `module` would break?

pmmmwh:
> You're talking about a global `module` variable as part of the commonjs standard, or any local variable named `module` would break?

As far as I know, any module-level declaration of `module` would break. I haven't tested local variables, maybe this is fixable if we change how we would search for exports, but no guarantees.

pmmmwh:
A change is now in place - exports failures will now warn in console instead of throwing.

kryops:
I ran into this via https://github.com/callstack/linaria/issues/891.
[...]

PR #726, opened 2023-02-10
match against moduleData.request

nicholasio:
matching against `moduleData.request` would allow folks to exclude virtual modules generated by other loaders (like linaria). For my use-case I can now do this:

```js
new ReactRefreshWebpackPlugin({
	exclude: [/node_module/, /outputCssLoader\.js/],
}),
```

To prevent react-refresh from being injected into CSS virtual modules (generated by linaria).

See https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/245#issuecomment-1426108711 for more context.

### DIFF
--- a/lib/utils/injectRefreshLoader.js
+++ b/lib/utils/injectRefreshLoader.js
@@ -28,7 +28,6 @@ function injectRefreshLoader(moduleData, injectOptions) {
   if (
     // Include and exclude user-specified files
     match(moduleData.matchResource || moduleData.resource) &&
-    match(moduleData.request) &&
     // Exclude files referenced as assets
     !moduleData.type.includes('asset') &&
     // Skip react-refresh and the plugin's runtime utils to prevent self-referencing -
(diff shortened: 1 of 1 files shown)


nicholasio:
Looks like this little change broke most of the conformance tests. Looking into it.

nicholasio:
@pmmmwh 👋 Any thoughts on this? If this doesn't feel like the right solution I'm happy to submit a separate PR to help address the issue.

Thanks!

lib/utils/injectRefreshLoader.js
```diff
@@ -38,6 +38,13 @@ function injectRefreshLoader(moduleData, injectOptions) {
     // Check to prevent double injection
     !moduleData.loaders.find(({ loader }) => loader === resolvedLoader)
   ) {
+    // if matchResource is set it means another loader is
+    // dynamically generating a module so let's look into moduleData.request
+    // to give users the opportunity to exclude them from instrumentation
+    if (moduleData.matchResource && !match(moduleData.request)) {
+      return moduleData;
+    }
```
pmmmwh, line 46: I think we can collapse this into the big if statement above?

nicholasio: IMO readability is easier by doing it outside since we only want to match against `moduleData.request` if `moduleData.matchResource` is set but maybe my brain is just not working well right now lol

pmmmwh: I guess another way to do this is to break down the if statement and do a bunch of early returns 🤔

PR #753, opened 2023-06-21
fix: correct packageType caching

DarkGenius:
Faced the problem of incorrect module type detection. In some cases, a global variable may contain a value for a package other than the package whose file is currently being processed. Adding `this.context` as a caching key fixes the bug


pmmmwh:
Tests are failing 🤔

PR #766, opened 2023-08-09
fix: Avoid unnecessary memory leaks due to prevExports

naruaway:
When fast-refreshing the following code:
```tsx
const DATA = Array.from({ length: 100000 }, (_, i) => Math.random());

export const App = () => {
  return (
    <div>
      <div>REWRITE_HERE</div>
      <div>{DATA.length}</div>
    </div>
  );
};
```
each refresh will not purge the previous DATA since we chain `prevExports`, which references `App`, which closes over `DATA`.  For reproducible example and more detailed explanation, please check https://github.com/naruaway-sandbox/fast-refresh-hmr-memory-leak-demo

Historical list of DATA is retained like the following via prevExports:

prev-exports-chain-leak

This PR breaks the chain by only passing absolutely necessary data across fast-refreshes, which is `signature: string[]` and `isReactRefreshBoundary: boolean`. So it will be impossible to accidentally form this kind of reference chain.

Note that I also tried to fix the same issue in Next.js https://github.com/vercel/next.js/pull/53797

## Verification

I verified using this PR with  https://github.com/naruaway-sandbox/fast-refresh-hmr-memory-leak-demo and confirmed that Fast Refresh itself is working well while not causing the memory leak

### DIFF
--- a/lib/runtime/RefreshUtils.js
+++ b/lib/runtime/RefreshUtils.js
@@ -63,6 +63,21 @@ function getReactRefreshBoundarySignature(moduleExports) {
 }
 
 /**
+ * Creates a data object to be retained across refreshes.
+ * This object should not transtively reference previous exports,
+ * which can form infinite chain of objects across refreshes, which can pressure RAM.
+ *
+ * @param {*} moduleExports A Webpack module exports object.
+ * @returns {*} A React refresh boundary signature array.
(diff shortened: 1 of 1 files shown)


pmmmwh (approved):
Nice catch, thank you!

PR #768, opened 2023-08-14
fix: get module system scoped to context

pmmmwh:
Fixes #753

### DIFF
--- a/loader/index.js
+++ b/loader/index.js
@@ -55,13 +55,18 @@ function ReactRefreshLoader(source, inputSourceMap, meta) {
   };
 
   /**
-   * @this {import('webpack').loader.LoaderContext}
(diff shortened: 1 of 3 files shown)


qnighy:
For those who encountered `require is not defined` issues after upgrading this package above v0.5.11: try upgrading webpack above v5.85.0 as the change introduced in this PR assumes the API introduced in https://github.com/webpack/webpack/pull/17281

(Not a bug report to this plugin itself. Thank you for the great plugin!)

Jinjiang:
@pmmmwh I hit an error earlier today with `webpack@5.84.1`. It's always cjs even the file is inside a node module with `type: module` in its `package.json`. I guess that's because older version of webpack doesn't have `this.environment` in the loader context. Now to work with that I have to downgrade to v0.5.9.

Considering the maximum compatibility, shall we check `this.environment` with an option? Or skip checking it with an option?

Thanks.

PR #786, opened 2023-11-28
To handle the routes on the client

dplopezsioux:
To handle the routes on the client, if the user forces the URL on the client side, Express will send the React app to the front-end, and then the application can handle the route in the front-end.


dplopezsioux:
I am using the example configuration, but with react-router, it does not work for me if I force the route in the browser because express only sends the application to the URL "/". I added that configuration to send the HTML to all the URLs so that React can manage the routes in the interface.

pmmmwh:
Thanks!

PR #834, opened 2024-04-10
Updates loader-utils to 3.2.1 due to Security Vulnerability

boroth:
Been resolving some Dependabot alerts in another repo, and ran into an issue with `react-refresh-webpack-plugin` needing to be updated to the newer version of `loader-utils`. This just updates the loader-utils package, but I'm also open to doing some additional yarn updating/auditing if people are interested in that.

Tests seem to be passing, just need to run a build and see if it's all good to go 👍🏾

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "error-stack-parser": "^2.0.6",
     "find-up": "^5.0.0",
     "html-entities": "^2.1.0",
-    "loader-utils": "^2.0.4",
+    "loader-utils": "^3.2.1",
     "schema-utils": "^3.0.0",
     "source-map": "^0.7.3"
   },
--- a/loader/index.js
+++ b/loader/index.js
@@ -5,7 +5,7 @@
 const originalFetch = global.fetch;
 delete global.fetch;
 
-const { getOptions } = require('loader-utils');
+const { loaderContext } = require('loader-utils');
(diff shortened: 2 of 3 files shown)


boroth:
I guess this fix may only work on Webpack 5, not sure what the standard way of handling that is in this project.

package.json
```diff
@@ -62,7 +62,7 @@
     "error-stack-parser": "^2.0.6",
     "find-up": "^5.0.0",
     "html-entities": "^2.1.0",
-    "loader-utils": "^2.0.4",
+    "loader-utils": "^3.2.1",
```
pmmmwh, line 65: Do we still consume `loader-utils` directly? If not I'm leaning towards removing this as a dependency to let downstream decide on what version is best for them.

pmmmwh:
> I guess this fix may only work on Webpack 5, not sure what the standard way of handling that is in this project.
> 
> I wouldn't be opposed to trying to get `react-refresh-webpack-plugin` to a `1.0.0` version with Node 16+ and Webpack 5+ if y'all are up for it.

I'll look a bit into if it's possible to have both supported and resolve the vulnerability. In any case I don't think we're _really_ using the code path that is reported, but it's nonetheless good to be on a version where this doesn't warn.

pmmmwh:
Will be fixed in `v0.6.0`.

Issue #742, opened 2023-03-28
Does not use `webSocketURL`


ChocolateLoverRaj:
Here is the docs for the option: https://webpack.js.org/configuration/dev-server/#websocketurl
This plugin just sends a request to `/ws`, even if a different path, hostname, or port is specified.
Link to reproduce: https://github.com/ChocolateLoverRaj/webpack-dev-server-bug/tree/react-refresh#readme

pmmmwh:
Support for this option have not been added, but should be. I'll try to find time to work on this.

Codex101:
I'm pretty sure this is why I keep getting an error with websocket connection failure. 
```
websocket connection to 'ws://localhost/ws' failed: Error during Websocket handshake: Unspected response code: 404
[webpack-dev-server] (error message)
[webpack-dev-server] Trying to reconnect... 

Constantly repeats
```

Codex101:
@pmmmwh Do you know where the plugin sends a the websocket connection request to webpack-dev-server, using the default `/ws` url? Trying to suppress this error on local. I cannot find where the ws requests get sent off.

ChocolateLoverRaj:
> @pmmmwh Do you know where the plugin sends a the websocket connection request to webpack-dev-server, using the default `/ws` url? Trying to suppress this error on local. I cannot find where the ws requests get sent off.

I did a text search on GitHub for "ws" and I found this: https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/22ace4ce61d9129d974071ffa4795d04121c60e7/sockets/utils/getSocketUrlParts.js#L81

Codex101:
@ChocolateLoverRaj The urlParts.pathname actually does pick up the webpack websocketUrl config value. 

eg, hardcoding the urlParts.pathname into my desired format doesn't work. 
```
  if (!urlParts.pathname) {
    if (metadata.version === 4) {
      // This is hard-coded in WDS v4
      urlParts.pathname = 'dist/ws';
    } else {
      // This is hard-coded in WDS v3
      urlParts.pathname = '/sockjs-node';
    }
  }
```

It's very weird, the WDSSocket.js does create a connection with the correct websocketUrl value from webpack, 'ws://localhost:1212/dist/ws`. However, the react refresh plugin just sends websocket connection requests to 'ws://localhost/ws`, same error my first comment.

pmmmwh:
[...]

Issue #728, opened 2023-02-17
HMR gathering information from my URL and hitting my backend instead no dev server


juniorgarcia:
Hello!

I have a backend for frontend running on port 3000. Is it possible to configure HMR in a way that it works correctly in those situations so that I don't get request errors when react refresh tries to gather information about refreshing my app?

I get this error because of the request hitting my express server:
![image](https://user-images.githubusercontent.com/16159850/219800877-9ea72ee3-8a7b-4059-8d3b-5cad956b349d.png)

The request URL is http://localhost:3000/ws/337/tt0fq4rk/jsonp?c=_jp.ant25bi and react refresh is one of the initiators of this request.

The response for this request is actually the main page my backend for frontend powered by express delivers.

Thank you!

PR #835, opened 2024-04-16
Do not override sockProtocol from config

darkexone:
Hey, I came across a case where `sockProtocol` is incorrectly overwritten. **My application does not use https, but is served through a proxy that uses https. I have https disabled in the webpack configuration**, but the protocol read from the window is https. **I tried passing the `sockProtocol` parameter, but found that wss is used all the time**, resulting in itegration with overlay not working. I believe that the config passed by the developer (a `resourceQuery`) should take priority over a failsafe configuration.

### DIFF
--- a/sockets/utils/getSocketUrlParts.js
+++ b/sockets/utils/getSocketUrlParts.js
@@ -103,10 +103,11 @@ function getSocketUrlParts(resourceQuery, metadata) {
   // We only re-assign `protocol` when `protocol` is unavailable,
   // or if `hostname` is available and is empty,
   // since otherwise we risk creating an invalid URL.
-  // We also do this when 'https' is used as it mandates the use of secure sockets.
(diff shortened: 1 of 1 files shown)


pmmmwh:
> A better idea would be to make a mechanism for reading the webSocketURL from the Webpack configuration (as suggested in https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/742 and https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/728), because this fix will help only for part of the problems with connecting to the wrong sockets. Until this mechanism is implemented, we should prefer the plugin configuration over the values read from the path.

With the later versions of WDS v4 and WDS v5 we can actually re-use the "created" server so we no-longer need to parse any of their config. It'll probably come in the next version though.

pmmmwh (approved):

pmmmwh:
Thanks for the PR!

Issue #824, opened 2024-03-25
Issues with peer dependencies found


rkmsnc:
Hi,
There is a WARN  Issues with peer dependencies found for webpack-dev-server 5.x version 
└─┬ @pmmmwh/react-refresh-webpack-plugin 0.5.11
  └── ✕ unmet peer webpack-dev-server@"3.x || 4.x": found 5.0.4

![Screenshot_20240326-000029](https://github.com/pmmmwh/react-refresh-webpack-plugin/assets/46848877/df67978b-1118-47e5-b6ec-ec9a9cd24264)

Thanks

nyngwang:
It had been reported, see #802.

Issue #800, opened 2024-02-13
Expand webpack-dev-server dev dependency to include the newes semver 5.x


Towerss:
Hi guys,

There is a new version of webpack-dev-server with a semver 5.0.0

This change would make it possible to update webpack-dev-server to the latest version 5.0.0 without needing to force the update.

It would also allow users to test with the latest version of webpack-dev-server in development. I installed in 2 projects bundled with webpack 5 and React 18, and I have have not issues with hot reload.

Issue #787, opened 2023-12-05
npm ERR! ERESOLVE could not resolve


glinkap:
![image](https://github.com/pmmmwh/react-refresh-webpack-plugin/assets/9206864/49ceb0a9-e664-4afa-8919-efc11d041c19)

binarykitchen:
https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/802 might fix this

PR #836, opened 2024-04-22
Allow v5 of `webpack-dev-server` as peer dependency

mxdvl:
[Release notes v5.0.0](https://github.com/webpack/webpack-dev-server/releases/tag/v5.0.0) along with [the migration guide](https://github.com/webpack/webpack-dev-server/blob/master/migration-v5.md).

- The Node version of the consumer is not a concern of this package
- The `webpack` and `webpack-cli` of the consumer are not a concern of this package
- The `http` & `https` options are not used internally

Should we also bump the local dev dependency on `webpack-dev-server` to v5?


pmmmwh:
> Should we also bump the local dev dependency on webpack-dev-server to v5?

I think we should, just to make sure tests pass. If that can be done I'm happy to cut this release first then work towards moving completely over to v5 👍🏻

pmmmwh:
Fixes #787 
Fixes #800 
Fixes #824

mxdvl:
@pmmmwh I’ve tried and failed to make CI pass:

- the minimum Node version is 18.12, and it seems your current puppeteer image is 18.9, so that fails for [webpack 4](https://app.circleci.com/pipelines/github/pmmmwh/react-refresh-webpack-plugin/1571/workflows/015ec0ce-04a9-457a-a59b-cd887a9a301b/jobs/11956?invite=true#step-0-346_72) and [webpack 5](https://app.circleci.com/pipelines/github/pmmmwh/react-refresh-webpack-plugin/1571/workflows/015ec0ce-04a9-457a-a59b-cd887a9a301b/jobs/11960)
- the test server is also failing to run… I am experiencing the same thing locally and feel like I am not familiar enough with the project to make this work

